PR #24, opened 2021-06-06
linter fixes

johnrees:
fixes some warnings that eslint shows when running the server

### DIFF
--- a/js/packages/web/src/views/artists/index.tsx
+++ b/js/packages/web/src/views/artists/index.tsx
@@ -1,18 +1,13 @@
-import React, { useState } from 'react';
-import { ArtCard } from '../../components/ArtCard';
-import { Layout, Row, Col, Tabs } from 'antd';
+import { Col, Layout } from 'antd';
+import React from 'react';
 import Masonry from 'react-masonry-css';
 import { Link } from 'react-router-dom';
-import { useUserArts } from '../../hooks';
-import { useMeta } from '../../contexts';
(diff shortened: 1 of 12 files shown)


js/packages/web/src/views/artists/index.tsx
```diff
@@ -1,18 +1,13 @@
...
+import { Col, Layout } from 'antd';
+import React from 'react';
 import Masonry from 'react-masonry-css';
 import { Link } from 'react-router-dom';
-import { useUserArts } from '../../hooks';
-import { useMeta } from '../../contexts';
 import { ArtistCard } from '../../components/ArtistCard';
-
-const { TabPane } = Tabs;
+import { useMeta } from '../../contexts';
```
johnrees, line 6: I'd be happy to do this for all files (remove unused imports + sort in a predictable order) but I think the size of the PR could be a bit annoying 😅  especially if people are currently working on branches e.g.
Screenshot 2021-06-06 at 11 09 12 AM

I do think it'd be good to run prettier and enforce linting on the whole project though, any thoughts?

johnrees:
+ the commit message `use colons in interface` should say semi-colons 🤦‍♂️

js/packages/web/src/views/artists/index.tsx
bartosz-lipinski: This is awesome, thanks a lot. That was definitely on my list ❤️

PR #55, opened 2021-06-17
Re-add testing tools to the build path

### LINKED ISSUE #54: Re-add the testing tools to the build path
All 4 programs (metaplex, auction, token-vault, token-metadata) also include a test CLI tool, but these have been removed from the build path due to some version mismatches with spl-token. These aren't an issue anymore so they can be re-added to the build path and used as a different way of learning how the contracts work.

praskoson:
Added the testing tools back to the build path. 
Also corrected the default RPC URLs, as the outdated ones were being used.

Fix #54

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -4,6 +4,10 @@ members = [
   "metaplex/program",
   "token-vault/program",
   "token-metadata/program",
+  "auction/cli",
+  "metaplex/test",
+  "token-valut/test",
+  "token-metadata/test",
 ]
 exclude = [
 ]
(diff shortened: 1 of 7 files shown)


bhgames:
We recently moved Vault to using a Key enum...the scripts haven't been updated to do that so this won't build. Can you make that update here?

bhgames (changes requested):
Needs fixes to scripts to have latest Key enum changes for Vault contract

praskoson:
I updated it, mind taking a look @bhgames?

bhgames:
This is gonna sound stupid, but I swear you're almost there :)

image

Just re-arrange the imports to get rustfmt to pass. We already aren't passing clippy, for which I feel guilty, but at least our vscode editors are keeping rustfmt auto-passing. Let's not fail that one too :D

praskoson:
I have to admit I didn't even know of rustfmt until now x)
Anyways, added those changes in the last commit.

bhgames (approved):

PR #84, opened 2021-06-28
Remove number "0" being rendered unintentionally

israelidanny:
`canCreate && (<Link...)` evaluates to a fals-ey value and renders as the number "0". The proper way to achieve not showing anything is via a ternary operator.

### DIFF
--- a/js/packages/web/src/components/AppBar/index.tsx
+++ b/js/packages/web/src/components/AppBar/index.tsx
@@ -26,9 +26,9 @@ const UserActions = () => {
       {/* <Link to={`#`}>
         <Button className="app-btn">Bids</Button>
       </Link> */}
-      {canCreate && (<Link to={`/art/create`}>
+      {canCreate ? (<Link to={`/art/create`}>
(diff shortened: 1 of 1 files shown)


israelidanny:
Here's an example of the unintended behavior
Screen Shot 2021-06-28 at 19 10 01

js/packages/web/src/components/AppBar/index.tsx
```diff
@@ -26,9 +26,9 @@ const UserActions = () => {
       {/* <Link to={`#`}>
         <Button className="app-btn">Bids</Button>
       </Link> */}
-      {canCreate && (<Link to={`/art/create`}>
+      {canCreate ? (<Link to={`/art/create`}>
```
bartosz-lipinski, line 29: Maybe just replace with !!canCreate

bartosz-lipinski:
> Here's an example of the unintended behavior
> Screen Shot 2021-06-28 at 19 10 01

Great catch

PR #96, opened 2021-06-29
Add dockerization for the build process

asghaier76:
Added a Dockerfile that will perform the entire build for prod process.
The Docker image also uses nginx to serve the frontend


bartosz-lipinski:
Thank you

Issue #100, opened 2021-06-29
Add 'hours' options to Auction time length 


GHesericsu:
Add 'hours' option to Auction time length when a user creates an auction so the user does not need to calculate how many minutes there are in 24 hours.

bhgames:
We dont have any plans to do this ourselves but would welcome help from the community.

nathanramli:
I think this issue can be closed now

PR #105, opened 2021-07-02
Add minutes, hours, days options on time input

nathanramli:
Add #100
- Add **minutes, hours, days** options on auction duration and gap time input.

I believe many auctions are open for more than 1 days. Using input in minutes is not effective.
Thanks.


bhgames:
added @bartosz-lipinski to take a look

bartosz-lipinski:
Awesome thanks a lot I will take a look tomorrow

PR #116, opened 2021-07-09
Discard this PR

PigeonSolToken:
Title says it all

### DIFF
--- a/js/packages/web/src/views/home/index.tsx
+++ b/js/packages/web/src/views/home/index.tsx
@@ -1,177 +1,1101 @@
-import React, { useState, useMemo } from 'react';
-import { Layout, Row, Col, Tabs, Button } from 'antd';
-import Masonry from 'react-masonry-css';
+import React, { useEffect, useState, useCallback } from 'react';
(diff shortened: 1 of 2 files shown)


js/packages/web/src/views/home/index.tsx
```diff
@@ -47,14 +54,19 @@ export const HomeView = () => {
   const liveAuctions = auctions
   .sort((a, b) => a.auction.info.endedAt?.sub(b.auction.info.endedAt || new BN(0)).toNumber() || 0);
 
+  const items =
+    activeKey === LiveAuctionViewState.All
+      ? liveAuctions
```
bhgames, line 59: So this only works for auctions that are currently live, not auctions that have finished, and it looks like the Live and Ended auctions were originally supposed to be tabs but now the tab functionality has broken down and they are just being displayed vertically one after each other on the page. Let's fix that and then make sure this is a new tab that displays. I'm not even sure this works properly. This particular tab should also have no limit of 10 or whatever that Ended Auctions seems to have.

Issue #103, opened 2021-07-01
Sanitize file names of cover image


sol-survivor:
The cover image of files with spaces don't load properly during NFT creation.
This can be fixed by modifying line 495 the artCreate/index.tsx
```
      const file = files.find(f => f.name.replace(/ /g, '-') === attributes.image);
```

However, after upload the cover image is not found on Arweave (404 error).

PigeonSolToken:
just remove all **cleanName** calls from the file artCreate/index.tsx. No need to use replace function.

bhgames:
@flappy-07 @sol-survivor this still happening for you ?

kemargrant:
> @flappy-07 @sol-survivor this still happening for you ?

I stopped using files with spaces in their name

sol-survivor:
> @flappy-07 @sol-survivor this still happening for you ?

I haven't tried loading an image with a space recently.

PR #117, opened 2021-07-09
Wider Browser Version Support for NFT Upload

kespinola:
### Issue

replaceAll is only available on Chrome 85+. The workstation is currently on Chrome 83 and the replaceAll method is undefined on string.

### Solution
Switch to replace with g flag which is the same but has greater browser support.


PigeonSolToken:
actually we don't even need this function, if we use this: https://github.com/metaplex-foundation/metaplex/issues/103#issuecomment-873452061. Cleanname causes the files having spaces to be not uploaded properly. Will create a PR

PR #120, opened 2021-07-12
add DRACO loader

Arrowana:
DRACO is a common data compression algorithm for GLTF.
I found out my artist was providing me models in this format, they were throwing
`Error: THREE.GLTFLoader: No DRACOLoader instance provided.`
In metaplex, but worked fine in phantom.

How about adding it to the viewer? It automatically picks the js implementation when webassembly isn't supported.
It loads it from file so I was not sure where should I put it, I went the lazy way and added it to public. Please advise if there is a better way.

This works in my store.


Arrowana:
I asked phantom what they are using and they said https://modelviewer.dev/
Should we move to this bundled viewer?

bhgames:
Added @bartosz-lipinski as a reviewer, he'd know best, i'm no expert here

PR #121, opened 2021-07-13
Migrate from CRA to next.js

exromany:


gching:
Is your linter different? Seems like a lot of code changes are linting changes as well. Can you make sure the linter is the same configuration as the repos?

Question regarding moving to Next.js though - what will moving to Next.js provide for Metaplex? Is it purely for performance? I am assuming most of the logic is client side logic (let me know if I am mistaken) and server side rendering will not add any benefits. I don't think SEO will benefit either since it is all client side logic

bartosz-lipinski:
> Is your linter different? Seems like a lot of code changes are linting changes as well. Can you make sure the linter is in the same configuration as the repos?
> 
> Question regarding moving to Next.js though - what will moving to Next.js provide for Metaplex? Is it purely for performance? I am assuming most of the logic is client-side logic (let me know if I am mistaken) and server-side rendering will not add any benefits. I don't think SEO will benefit either since it is all client-side logic

Overall biggest complaint about metaplex on the client side right now is initial paint performance. Move to Next.Js is a long term attempt at addressing some of those with still allowing people to deploy to github pages. 
I think we should be able to do few things:
* Ability to easily integrate landing pages for projects that load fast
* eventually try to move initial render to server-side for the faster load (just need to figure out how to integrate getStaticProps with WebSocket subscriptions). Good example would 
* If we figure out how to query for metadata in a good way we should be able to optimize image rendering as well

bartosz-lipinski:
> Is your linter different? Seems like a lot of code changes are linting changes as well. Can you make sure the linter is in the same configuration as the repos?
> 
> Question regarding moving to Next.js though - what will moving to Next.js provide for Metaplex? Is it purely for performance? I am assuming most of the logic is client-side logic (let me know if I am mistaken) and server-side rendering will not add any benefits. I don't think SEO will benefit either since it is all client-side logic

Overall biggest complaint about metaplex on the client side right now is initial paint performance. Move to Next.Js is a long term attempt at addressing some of those with still allowing people to deploy to github pages. 
I think we should be able to do few things:
* Ability to easily integrate landing pages for projects that load fast
* eventually try to move initial render to server-side for the faster load (just need to figure out how to integrate getStaticProps with WebSocket subscriptions). Good example would 
* If we figure out how to query for metadata in a good way we should be able to optimize image rendering as well

js/packages/web/src/pages/_document.tsx
```diff
@@ -0,0 +1,76 @@
...
+            src="https://platform.twitter.com/widgets.js"
+            charSet="utf-8"
+          />
+          <script
+            type="text/javascript"
+            dangerouslySetInnerHTML={{
+              __html: `
+                (function () {
+                  var s = document.createElement("script");
+                  s.src = "https://stackpile.io/stack_162299.js"; s.async = true;
```
kevinrodriguez-io, line 66: Is this stackpile account going to be enabled for all sites/forks?
If not, then I think a good approach would be to leave this empty and add it to the docs on how to set up the store.

js/packages/web/src/pages/_document.tsx
```diff
@@ -0,0 +1,76 @@
...
+            type="image/png"
+            sizes="16x16"
+            href="/favicon-16x16.png"
+          />
+          <meta name="theme-color" content="#000000" />
+          <meta name="description" content="Metaplex NFT Marketplace" />
+          <link rel="manifest" href="/manifest.json" />
+          <link
+            rel="stylesheet"
+            href="//cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/css/all.min.css"
```
kevinrodriguez-io, line 47: If possible, it would be good to reduce attack layers by using the npm version of this instead of fetching from the cdn

js/packages/web/src/pages/_document.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  render() {
+    return (
+      <Html lang="en">
+        <Head>
+          <meta charSet="utf-8" />
+          <meta
+            name="loadforge-site-verification"
+            content="2056680d2883a8b910880d53b9cb2ebf16e7b8f91e169cceddce62c4c4ef8fe6240748c08c2e3769e554e12dafcd5bfc62028638e6524a0efd7d729efd762d42"
```
kevinrodriguez-io, line 22: I think this can also be removed from the source right?

kevinrodriguez-io (commented):
Hi guys, amazing job done with this PR and Metaplex overall. You rock 🚀

I wanted to leave some small notes on some external-script usage.

I'm aware this existed on the index.html file of the create-react-app version, but I think now it's going to be easier to add env variables or remove some not-needed scripts.

js/packages/web/src/pages/_document.tsx
bartosz-lipinski: created an issue to track it

js/packages/web/src/pages/_document.tsx
bartosz-lipinski: created an issue to track it

PR #126, opened 2021-07-15
[FIX] Show correct edition on the art & auction page

sultanpeyek:
Currently, users can't see which edition of the NFT that currently be auctioned. Also, there are no information about edition on art detail page.

So here are the results of the fixes that I propose.

Check the [live example here](https://metaplex-byplmn92t-sultanpeyek.vercel.app/#/auction/2EbfuhBoTJqMWQ2tWSbEg6ma43vUsvCBvn6mMobEZnAM) (change the connection to devnet)

### Edition Info on Art detail page
image

### Edition Info on Auction page
Multiple
image

Specific edition of the copy
image

Master (NFT 0)
image

### DIFF
--- a/js/packages/web/src/views/art/index.less
+++ b/js/packages/web/src/views/art/index.less
@@ -33,3 +33,9 @@
   font-weight: 600;
   font-size: 1.3rem;
 }
+
+.art-edition {
+  font-weight: 600;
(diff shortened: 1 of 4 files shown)


bartosz-lipinski:
❤️ Thank you

PR #128, opened 2021-07-16
[FEATURE] Implement Bonfida Twitter Handling

sultanpeyek:
Implement [Bonfida Name Service](https://www.npmjs.com/package/@bonfida/spl-name-service) (Twitter Handle) to show twitter username of the bidder. This feature gets good responses from the Solana eco [twitter](https://twitter.com/_Dean_Machine/status/1415905445768798210?s=20) since it could lit up the bidding wars.

### Screenshot Example
image

Feel free if you guys want to refactor the codes haha.

PS: I added `webpack-dev-server` on the package.json because I kept getting these error after adding `@bonfida/spl-name-service` package.
```
web: There might be a problem with the project dependency tree.
web: It is likely not a bug in Create React App, but something you need to fix locally.
web: The react-scripts package provided by Create React App requires a dependency:
web:   "webpack-dev-server": "3.11.0"
```

### DIFF
--- a/js/packages/web/src/views/auction/index.tsx
+++ b/js/packages/web/src/views/auction/index.tsx
@@ -1,7 +1,8 @@
-import React, { useRef, useState } from 'react';
+import React, { useRef, useState, useEffect } from 'react';
 import { useParams } from 'react-router-dom';
 import { Row, Col, Button, Skeleton, Carousel } from 'antd';
 import { AuctionCard } from '../../components/AuctionCard';
+import { Connection, PublicKey } from '@solana/web3.js';
 import {
   AuctionView as Auction,
   AuctionViewItem,
@@ -19,6 +20,7 @@ import {
   Identicon,
   MetaplexModal,
   shortenAddress,
+  useConnection,
   useConnectionConfig,
   fromLamports,
   useMint,
@@ -26,6 +28,7 @@ import {
   AuctionState,
 } from '@oyster/common';
 import { MintInfo } from '@solana/spl-token';
+import { getHandleAndRegistryKey } from '@bonfida/spl-name-service';
 import useWindowDimensions from '../../utils/layout';
 import { CheckOutlined } from '@ant-design/icons';
 import { useMemo } from 'react';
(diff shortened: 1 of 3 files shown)


bartosz-lipinski:
very nice that was one of the items on my list. 🙇‍♂️ ❤️

js/packages/web/src/views/auction/index.tsx
```diff
@@ -202,6 +205,28 @@ const BidLine = (props: { bid: any; index: number; mint?: MintInfo, isCancelled?
...
+  // Get Twitter Handle from address
+  const connection = useConnection();
+  const [bidderTwitterHandle, setBidderTwitterHandle] = useState('');
+  useEffect(() => {
+    const getTwitterHandle = async (
+      connection: Connection,
+      bidder: PublicKey,
+    ): Promise<string | undefined> => {
+      try {
+        const [twitterHandle] = await getHandleAndRegistryKey(
```
bartosz-lipinski, line 217: does this one use getMultipleAccounts? I want to make sure we are not blowing up rate limits?

js/packages/web/src/views/auction/index.tsx
```diff
@@ -202,6 +205,28 @@ const BidLine = (props: { bid: any; index: number; mint?: MintInfo, isCancelled?
...
+      } catch (err) {
+        console.warn(`err`);
+        return undefined;
+      }
+    };
+    getTwitterHandle(connection, bidder);
+  }, [bidderTwitterHandle]);
+
   return (
     <Row
```
bartosz-lipinski, line 231: can you also pull twitter image?

bartosz-lipinski (approved):

js/packages/web/src/views/auction/index.tsx
sultanpeyek: Sorry I just read this, will check on that.

js/packages/web/src/views/auction/index.tsx
sultanpeyek: Seems that it's not doable just by using Bonfida, but I'll try to figure it out.
Thanks @bartosz-lipinski!

js/packages/web/src/views/auction/index.tsx
sultanpeyek: FYI @bartosz-lipinski 
https://github.com/solana-labs/solana-program-library/blob/master/name-service/js/src/twitter.ts#L245

bartosz-lipinski: Yeah that one is using individual getAccountInfo and doesn’t cache. We should query for all visible bids with single call and cache

PR #142, opened 2021-07-19
Fix webpack import and add wallet adapters to export

ChewingGlass:
Some minor changes that allow the Wum.bo codebase to be compatible with this codebase.


js/packages/common/src/contexts/accounts.tsx
```diff
@@ -146,10 +146,6 @@ export const cache = {
     parser?: AccountParser,
     isActive?: boolean | undefined | ((parsed: any) => boolean),
   ) => {
-    if (obj.data.length === 0) {
```
ChewingGlass, line 149: Removing this allows us to fetch and watch the SOL balance of a wallet, which may be empty.

PR #145, opened 2021-07-20
web: fix export & deploy

exromany:
fixed exporting static html for gh-pages

### DIFF
--- a/js/packages/web/next.config.js
+++ b/js/packages/web/next.config.js
@@ -5,6 +5,9 @@ const dotenvLoad = require('dotenv-load');
 
 dotenvLoad();
 
+const debug = process.env.NODE_ENV !== 'production';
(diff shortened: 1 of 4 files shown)


js/packages/web/next.config.js
```diff
@@ -5,6 +5,9 @@ const dotenvLoad = require('dotenv-load');
 
 dotenvLoad();
 
+const debug = process.env.NODE_ENV !== 'production';
+const assetPrefix = !debug ? '/metaplex/' : '';
```
bartosz-lipinski, line 9: can you move this to .env file

bartosz-lipinski (changes requested):

js/packages/web/next.config.js
exromany: ok, but you will have to specify the env before deploy building

PR #159, opened 2021-07-23
Add instant sale feature to Auction program

VadimGrozinok:


rust/auction/program/src/processor.rs
```diff
@@ -90,7 +90,7 @@ pub struct AuctionData {
     pub bid_state: BidState,
 }
 
-pub const MAX_AUCTION_DATA_EXTENDED_SIZE: usize = 8 + 9 + 2 + 200;
+pub const MAX_AUCTION_DATA_EXTENDED_SIZE: usize = 8 + 9 + 2 + 9 + 200;
```
bhgames, line 93: You need to use the existing padding, you can't make the accounts larger. So this would be + 9 + 191

rust/auction/program/src/processor.rs
```diff
@@ -102,6 +102,8 @@ pub struct AuctionDataExtended {
     // Unimplemented fields
     /// Tick size
     pub tick_size: Option<u64>,
+    /// instant sale price
+    pub instant_sale_price: Option<u64>,
```
bhgames, line 106: When adding a field to an existing struct you need to put it at the end, so it is backwards compatible with other structs like it. Place this after gap tick size percentage.

rust/auction/program/src/processor.rs
```diff
@@ -505,6 +575,18 @@ impl BidState {
...
                         }
+
+                        // Check if all the lots were sold with instant_sale_price
+                        if let Some(instant_sale_amount) = instant_sale_price {
+                            // bids.len() - max = index of the last winner bid
+                            if bids.len() >= *max
+                                && bids[bids.len() - *max].1 >= instant_sale_amount
+                            {
+                                msg!("All the lots were sold with instant_sale_price, auction is ended");
+                                *auction_sate = AuctionState::Ended;
```
bhgames, line 586: can we make this called auction_state, not auction_sate?

rust/auction/program/src/processor.rs
```diff
@@ -116,6 +118,33 @@ impl AuctionDataExtended {
...
+            Some(u64::from_le_bytes(*array_ref![data, idx, 8]))
+        } else {
+            None
+        }
+    }
+
+    fn find_instant_sale_beginning<'a>(data: &'a Ref<'a, &'a mut [u8]>) -> Option<usize> {
+        // total_uncancelled_bids + tick_size Option
+        let mut instant_sale_beginning = 8;
+
```
bhgames, line 133: Obviously given my above comment, this will change when you move the field

rust/auction/program/src/processor/cancel_bid.rs
```diff
@@ -175,10 +179,20 @@ pub fn cancel_bid(
...
+    let winner_bid_index = auction.is_winner(accounts.bidder.key);
+    if auction.ended(clock.unix_timestamp)? && winner_bid_index.is_some() {
         return Err(AuctionError::InvalidState.into());
     }
 
+    // Refuse to cancel if bidder set price above or equal instant_sale_price
+    if let Some(bid_index) = winner_bid_index {
+        if let Some(instant_sale_price) = auction_extended.instant_sale_price {
+            if auction.bid_state.amount(bid_index) >= instant_sale_price {
+                return Err(AuctionError::InvalidState.into());
```
bhgames, line 191: nice

rust/auction/program/src/processor/claim_bid.rs
```diff
@@ -57,13 +58,15 @@ fn parse_accounts<'a, 'b: 'a>(
         bidder_pot: next_account_info(account_iter)?,
         authority: next_account_info(account_iter)?,
         auction: next_account_info(account_iter)?,
+        auction_extended: next_account_info(account_iter)?,
```
bhgames, line 61: So this is a problem. When adding to an existing contract endpoint, you can't put an account in the middle of the expected set because you will break existing older client UIs. So in this instance what you can do is put it beneath token program, and make it optional. If it's not present in the next_account_info, then you wont check for instant sale-ness, and the caller must wait till the auction ends like normal. If they wish to claim early, they need to provide auction data extended.

rust/auction/program/src/processor/create_auction.rs
```diff
@@ -44,6 +44,8 @@ pub struct CreateAuctionArgs {
     pub tick_size: Option<u64>,
     /// Add a minimum percentage increase each bid must meet.
     pub gap_tick_size_percentage: Option<u8>,
+    /// Add a instant sale price.
+    pub instant_sale_price: Option<u64>,
```
bhgames, line 48: Again, changing an existing contract struct. This will break older UIs. The best way to go with this one is to actually create a new contract endpoint called create_auction_v2.rs that takes a modified struct and just calls the original with more parameters, something like this, so you have complete code re-use but one takes a slightly different struct.

rust/auction/program/src/processor/place_bid.rs
```diff
@@ -281,12 +281,21 @@ pub fn place_bid<'r, 'b: 'r>(
         .ok_or(AuctionError::NumericalOverflowError)?;
     auction_extended.serialize(&mut *accounts.auction_extended.data.borrow_mut())?;
 
+    let mut bid_price = args.amount;
+
+    if let Some(instant_sale_price) = auction_extended.instant_sale_price {
+        if args.amount > instant_sale_price {
+            msg!("Received amount is more than instant_sale_price so it was reduced to instant_sale_price - {:?}", instant_sale_price);
```
bhgames, line 288: nice

rust/metaplex/program/src/instruction.rs
```diff
@@ -143,21 +144,22 @@ pub enum MetaplexInstruction {
...
-    ///   9. `[signer optional]` Your Bidder account - Only needs to be signer if payer does not own
-    ///   10. `[signer]` Payer
-    ///   11. `[]` Token program
-    ///   12. `[]` Token Vault program
-    ///   13. `[]` Token metadata program
-    ///   14. `[]` Store
-    ///   15. `[]` System
-    ///   16. `[]` Rent sysvar
-    ///   17. `[writable]` Master Metadata account (pda of ['metadata', program id, Printing mint id]) - remember PDA is relative to token metadata program
+    ///   8. `[]` Auction extended
```
bhgames, line 147: Yah again, can't do this - you'll break older UIs. Got to put it at the end, or make a new end point.

rust/metaplex/program/src/instruction.rs
```diff
@@ -109,19 +109,20 @@ pub enum MetaplexInstruction {
...
-    ///   9. `[signer optional]` Your Bidder account - Only needs to be signer if payer does not own
-    ///   10. `[signer]` Payer
-    ///   11. `[]` Token program
-    ///   12. `[]` Token Vault program
-    ///   13. `[]` Token metadata program
-    ///   14. `[]` Store
-    ///   15. `[]` System
-    ///   16. `[]` Rent sysvar
-    ///   17. `[]` PDA-based Transfer authority to move the tokens from the store to the destination seed ['vault', program_id, vault key]
+    ///   8. `[]` Auction extended
```
bhgames, line 112: Yah again, can't do this - you'll break older UIs. Got to put it at the end, or make a new end point.

rust/metaplex/program/src/instruction.rs
```diff
@@ -182,20 +184,21 @@ pub enum MetaplexInstruction {
...
-    ///   9. `[signer optional/writable]` Your Bidder account - Only needs to be signer if payer does not own
-    ///   10. `[signer]` Payer
-    ///   11. `[]` Token program
-    ///   12. `[]` Token Vault program
-    ///   13. `[]` Token metadata program
-    ///   14. `[]` Store
-    ///   15. `[]` System
-    ///   16. `[]` Rent sysvar
-    ///   17. `[signer]` Transfer authority to move the payment in the auction's token_mint coin from the bidder account for the participation_fixed_price
+    ///   8. `[]` Auction extended
```
bhgames, line 187: Yah again, can't do this - you'll break older UIs. Got to put it at the end, or make a new end point.

rust/metaplex/program/src/instruction.rs
```diff
@@ -389,24 +392,25 @@ pub enum MetaplexInstruction {
...
-    ///   14. `[]` Store
-    ///   15. `[]` System
-    ///   16. `[]` Rent sysvar
-    ///   17. `[writable]` Prize tracking ticket (pda of ['metaplex', program id, auction manager key, metadata mint id])
-    ///   18. `[writable]` New Metadata key (pda of ['metadata', program id, mint id])
-    ///   19. `[writable]` New Edition (pda of ['metadata', program id, mint id, 'edition'])
-    ///   20. `[writable]` Master Edition of token in vault V2 (pda of ['metadata', program id, master metadata mint id, 'edition']) PDA is relative to token metadata.
-    ///   21. `[writable]` Mint of new token
-    ///   22. `[writable]` Edition pda to mark creation - will be checked for pre-existence. (pda of ['metadata', program id, master metadata mint id, 'edition', edition_number])
+    ///   8. `[]` Auction extended
```
bhgames, line 395: Yah again, can't do this - you'll break older UIs. Got to put it at the end, or make a new end point.

rust/metaplex/program/src/instruction.rs
```diff
@@ -454,29 +458,30 @@ pub enum MetaplexInstruction {
...
-    ///   9. `[]` Your Bidder account
-    ///   10. `[signer]` Payer
-    ///   11. `[]` Token program
-    ///   12. `[]` Token Vault program
-    ///   13. `[]` Token metadata program
-    ///   14. `[]` Store
-    ///   15. `[]` System
-    ///   16. `[]` Rent sysvar
-    ///   17. `[signer]` Transfer authority to move the payment in the auction's token_mint coin from the bidder account for the participation_fixed_price
+    ///   8. `[]` Auction extended
```
bhgames, line 461: Yah again, can't do this - you'll break older UIs. Got to put it at the end, or make a new end point.

bhgames (changes requested):
Overall, great first attempt here! I love how you've spent some time really thinking about how it all connects. I think you need to consider more though that there are people with forks of the JS code that are way out of date and that will continue to expect their sites to work. We have to build for them, too. Sometimes that means building new endpoints that call the same method with extra parameters, and sometimes it means just adding an account to the end of the iterator and treating it as an Option<&AccountInfo>. I tend to prefer the latter approach as it's less polluting to the Instruction enum but there are occasional cases where you can't get around it, like when changing the struct you're passing into the enum instruction. So revisit this code considering that angle of backwards compatibility and let's look at it again!

rust/auction/program/src/instruction.rs
```diff
@@ -30,10 +31,20 @@ pub enum AuctionInstruction {
     /// Create a new auction account bound to a resource, initially in a pending state.
     ///   0. `[signer]` The account creating the auction, which is authorised to make changes.
     ///   1. `[writable]` Uninitialized auction account.
-    ///   2. `[]` Rent sysvar
-    ///   3. `[]` System account
+    ///   2. `[writable]` Auction extended data account.
```
bhgames, line 34: Is this a mistake that this is still here?

rust/auction/program/src/instruction.rs
```diff
@@ -30,10 +31,20 @@ pub enum AuctionInstruction {
...
+    ///   3. `[]` Rent sysvar
+    ///   4. `[]` System account
     CreateAuction(CreateAuctionArgs),
 
+    /// Create a new auction account bound to a resource, initially in a pending state.
+    /// The only one difference with above instruction it's instant_sale_price parameter in CreateAuctionArgsV2
+    ///   0. `[signer]` The account creating the auction, which is authorised to make changes.
+    ///   1. `[writable]` Uninitialized auction account.
```
bhgames, line 42: By placing this here instead of at the end of the enum, you have now offset the contract commands for every command after it and supplanted the contract command that used to have this number(ClaimBid). This means when old UIs call this contract expecting to hit ClaimBid, they'll hit CreateAuctionV2 instead, and when they call anything else, they'll always be "off by one." You need to place any new contract commands at the end of the Instruction enum so as not to disrupt older clients, because the clients pass u8s to tell the program which command to use and these are just indexing into the Instruction enum.

rust/auction/program/src/processor.rs
```diff
@@ -505,6 +582,18 @@ impl BidState {
...
                         }
+
+                        // Check if all the lots were sold with instant_sale_price
+                        if let Some(instant_sale_amount) = instant_sale_price {
+                            // bids.len() - max = index of the last winner bid
+                            if bids.len() >= *max
+                                && bids[bids.len() - *max].1 >= instant_sale_amount
+                            {
+                                msg!("All the lots were sold with instant_sale_price, auction is ended");
+                                *auction_state = AuctionState::Ended;
```
bhgames, line 593: Thinking about this, I think instead of passing in a mutable auction state, which is kind of a violation of interfaces and the law of demeter, we should instead have a function called place_bid on auction, that calls place bid on this, and it does this extra if statement check at that level, because that's the appropriate place for a change to auction status, not inside this other struct. Then you can do self.status = AuctionState::Ended. See the other wrapper methods in Auction already present. Also serves additional side effect of doing the check for all types of auctions as opposed to just english!

rust/auction/program/src/processor/claim_bid.rs
```diff
@@ -120,14 +126,43 @@ pub fn claim_bid(
...
-    if auction.is_winner(accounts.bidder.key).is_none() {
+    let bid_index = auction.is_winner(accounts.bidder.key);
+    if bid_index.is_none() {
         msg!("User {:?} is not winner", accounts.bidder.key);
         return Err(AuctionError::InvalidState.into());
     }
 
-    // Auction must have ended.
+    let mut auction_extended: Option<AuctionDataExtended> = None;
+    if let Some(auction_extended_info) = accounts.auction_extended {
```
bhgames, line 136: Maybe instead of doing all these gymnastics to get at an Option<AuctionDataExtended>, when really you want to get an Option<u64>, why not actually do let mut instant_sale_price: Option<u64> = None, and then set it in here? After all, that's what you really want, not the AuctionDataExtended. Then it's much easier to check down there. Then the reader is not thinking to them selves, "why do I know for sure instant sale price is unwrappable?" without having to know what you were doing in this bulky if statement.

rust/metaplex/program/src/instruction.rs
```diff
@@ -123,6 +123,7 @@ pub enum MetaplexInstruction {
     ///   18. `[optional/writable]` Master edition (if Printing type of WinningConfig)
     ///   19. `[optional/writable]` Reservation list PDA ['metadata', program id, master edition key, 'reservation', auction manager key]
     ///        relative to token metadata program (if Printing type of WinningConfig)
+    ///   20. `[]` Auction extended
```
bhgames, line 126: I normally leave comment like PDA of ['auction', program id, resource, 'extended'] relative to auction contract next to this so people know how to derive it(I think that's the correct seed).

rust/metaplex/program/src/utils.rs
```diff
@@ -604,7 +606,19 @@ pub fn common_redeem_checks(
     }
 
     if AuctionData::get_state(auction_info)? != AuctionState::Ended {
-        return Err(MetaplexError::AuctionHasNotEnded.into());
+        if win_index.is_some() && auction_extended_info.is_some() {
```
bhgames, line 609: To keep it clean let's abstract this into it's own method. Something like assert_auction_is_ended_or_this_is_valid_instant_sale()?

bhgames: In the case of open edition auctions, win_index may not be some. It will be none. That should still work just fine. Have you considered the OpenEdition auction type? Because that's the case of selling an infinite number of items at 5SOL each :)

rust/metaplex/program/src/utils.rs
```diff
@@ -604,7 +606,19 @@ pub fn common_redeem_checks(
...
+            if let Some(instant_sale_price) = AuctionDataExtended::get_instant_sale_price(&auction_extended_info.unwrap().data.borrow()) {
+                // we can safely do unwrap here such as existing win_index proves that we can get winner_bid_price
+                let winner_bid_price = AuctionData::get_winner_bid_amount_at(auction_info, win_index.unwrap()).unwrap();
+                if winner_bid_price < instant_sale_price {
```
bhgames, line 613: For the OpenEdition auction you can get the bid price from the last_bid field on the BidderMetadata which is present here if overwrite_win_index.is_none(). Might actually be easier than using get_winner_bid_amount_at, which is specific to winners.

bhgames (changes requested):
OK overall, another great attempt. Something it doesn't appeared you have considered is the ramifications of using an instant sale settings with the OpenEdition enum type. The OpenEdition auction doesnt use the bidders array and max winners is set to 0. People can still bid any amount they like though - now with instant sale, it'll be fixed at hard top cap of instant sale price, and they can't cancel their bids once they've bid that. You need to think about what happens when there is no bid in a bidding array to look up prices by in those scenarios.

Alright, round 2!

rust/auction/program/src/instruction.rs
vecheslav: the account is used in the instruction itself. i think the comment looks correct

ysavchenko (commented):
Just a small comment regarding architectural documentation.

ARCHITECTURE.md
```diff
@@ -456,6 +456,8 @@ pub struct AuctionDataExtended {
     pub tick_size: Option<u64>,
     /// gap_tick_size_percentage - two decimal points
     pub gap_tick_size_percentage: Option<u8>,
+    /// auction name
+    pub name: [u8; 32],
```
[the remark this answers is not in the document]
bhgames: Probably dont want to update here is the master version of this is in a notion. This is a copy that gets overwritten perioidically. https://www.notion.so/Metaplex-Developer-Guide-afefbc19841744c28587ab948a08cfac < Make update there :)

rust/auction/program/src/processor.rs
```diff
@@ -606,6 +706,17 @@ impl BidState {
             BidState::OpenEdition { bids, max } => None,
         }
     }
+
+    pub fn has_instant_bid(&self, instant_sale_amount: u64) -> bool {
```
bhgames, line 710: i was confused by this naming - minor nit - rename it to "lowest_winning_bid_is_instant_bid_price"

rust/metaplex/program/src/utils.rs
```diff
@@ -176,6 +177,39 @@ pub fn assert_authority_correct(
...
+    let instant_sale_price = auction_extended_info
+        .and_then(|info| AuctionDataExtended::get_instant_sale_price(&info.data.borrow()));
+
+    match instant_sale_price {
+        Some(instant_sale_price) => {
+            let winner_bid_price = if let Some(win_index) = win_index {
+                AuctionData::get_winner_bid_amount_at(auction_info, win_index).unwrap()
+            } else {
+                // Possible case in an open auction
+                BidderMetadata::from_account_info(bidder_metadata_info)?.last_bid
```
bhgames, line 200: Nice, love to see the edge case handling.

bhgames (commented):
Alright this looks great. Had some minor naming nits but this is ready to go to UI development. I would not say we can merge it until it's got a UI on top to prove that it works in an integration sense, but the Rust looks great and looks like it'll work just fine.

rust/auction/program/src/instruction.rs
```diff
@@ -30,8 +31,9 @@ pub enum AuctionInstruction {
     /// Create a new auction account bound to a resource, initially in a pending state.
     ///   0. `[signer]` The account creating the auction, which is authorised to make changes.
     ///   1. `[writable]` Uninitialized auction account.
-    ///   2. `[]` Rent sysvar
-    ///   3. `[]` System account
+    ///   2. `[writable]` Auction extended data account (pda relative to auction of ['auction', program id, vault key, 'extended']).
```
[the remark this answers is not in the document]
ysavchenko: Not really, it was there, but missing in the documentation, so this change was for simply adding it to the comments before the instruction.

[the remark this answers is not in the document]
bhgames: ahhh

rust/metaplex/program/src/utils.rs
vecheslav: Sorry, which cases do u mean? if understand correctly, all we need to know `winner_bid_price < instant_sale_price`

bhgames: Oh I wasn't saying you were missing anything. It was a compliment.

bhgames:
Just fyi @ysavchenko all changes to Architecture.md get periodically wiped from the master that exists on Notion.

ysavchenko:
> Just fyi @ysavchenko all changes to Architecture.md get periodically wiped from the master that exists on Notion.

Got it.
Also, @bhgames I think we're done with this one, all comments seems to be resolved. Can you please take another look at it and merge if everything is ok?

rust/metaplex/program/src/processor/end_auction.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    if auction_manager.auction != *auction_info.key {
+        return Err(MetaplexError::AuctionManagerAuctionMismatch.into());
+    }
+
+    if store.auction_program != *auction_program_info.key {
+        return Err(MetaplexError::AuctionManagerAuctionProgramMismatch.into());
+    }
+
+    if auction_manager.state.status != AuctionManagerStatus::Validated {
```
bhgames, line 86: Auction manager would be in running state

rust/metaplex/program/src/processor/end_auction.rs
```diff
@@ -0,0 +1,124 @@
...
+        return Err(MetaplexError::AuctionManagerMustBeValidated.into());
+    }
+
+    let seeds = &[PREFIX.as_bytes(), &auction_manager.auction.as_ref()];
+    let (_, bump_seed) = Pubkey::find_program_address(seeds, &program_id);
```
bhgames, line 91: use assert_derivation helper

rust/metaplex/program/src/processor/end_auction.rs
```diff
@@ -0,0 +1,124 @@
...
+        auction_program_info.clone(),
+        auction_manager_info.clone(),
+        auction_info.clone(),
+        clock_info.clone(),
+        auction_manager.vault,
+        args.reveal,
+        authority_seeds,
+    )?;
+
+    if auction_data_extended.instant_sale_price.is_some() {
```
bhgames, line 108: Why is this true?

rust/metaplex/program/src/processor/end_auction.rs
bhgames: It'll also serve as checking that auction manager is truly derived from auction as it should be

rust/metaplex/program/src/processor/end_auction.rs
```diff
@@ -0,0 +1,124 @@
...
+    let auction = AuctionData::from_account_info(auction_info)?;
+    let auction_data_extended = AuctionDataExtended::from_account_info(auction_data_extended_info)?;
+    let store = Store::from_account_info(store_info)?;
+
+    if auction.authority != *auction_manager_info.key {
+        return Err(MetaplexError::AuctionAuthorityMismatch.into());
+    }
+
+    assert_authority_correct(&auction_manager, authority_info)?;
+    assert_owned_by(auction_info, &store.auction_program)?;
```
bhgames, line 70: forgot to check ownership of extended

bhgames (changes requested):
Just some minor changes needed!

rust/metaplex/program/src/processor/end_auction.rs
VadimGrozinok: It's true when user created unlimited(in time) auction with instant_sale feature
Or I didn't get the question ?

PR #160, opened 2021-07-24
Updates

edcfn:


bartosz-lipinski:
what are these updates of?

bhgames:
Can we get some color on what these updates are? It seems like changing a bunch of static files.

bartosz-lipinski:
Closing since there are no changes and no replies.

PR #166, opened 2021-07-25
Added name to the auction data

### LINKED ISSUE #112: Add a name field to auction so users can customize their auction
Will have to be in AuctionDataExtended due to the borsh stack limit issue. Unfortunate necessity.

ysavchenko:
Closes #112

### DIFF
--- a/js/packages/common/src/actions/auction.ts
+++ b/js/packages/common/src/actions/auction.ts
@@ -15,7 +15,7 @@ import { findProgramAddress } from '../utils';
 export const AUCTION_PREFIX = 'auction';
 export const METADATA = 'metadata';
 export const EXTENDED = 'extended';
-export const MAX_AUCTION_DATA_EXTENDED_SIZE = 8 + 9 + 2 + 200;
+export const MAX_AUCTION_DATA_EXTENDED_SIZE = 8 + 9 + 2 + 32 + 168;
 
 export enum AuctionState {
   Created = 0,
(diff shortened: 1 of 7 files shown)


js/packages/common/src/actions/auction.ts
```diff
@@ -177,15 +177,18 @@ export class AuctionDataExtended {
   totalUncancelledBids: BN;
   tickSize: BN | null;
   gapTickSizePercentage: number | null;
+  name: Uint8Array;
```
bhgames, line 180: oh I didnt realize this was a type we could use.

rust/auction/program/src/processor/create_auction.rs
```diff
@@ -44,6 +44,8 @@ pub struct CreateAuctionArgs {
     pub tick_size: Option<u64>,
     /// Add a minimum percentage increase each bid must meet.
     pub gap_tick_size_percentage: Option<u8>,
+    /// Auction name
+    pub name: [u8; 32],
```
bhgames, line 48: same thing as I told Vadim, you break older clients when you do this. You need to create a new create auction endpoint to support this new struct, and basically call the same code - otherwise older clients will get a borshio error when they hit this and they won't know why! Also, name should be an Option tag on the model, so that it can optionally not be set if using the older endpoint.

bhgames (changes requested):
We're going to have to be careful here. With you and vadim both making a new create auction end point, it makes sense to merge both branches together into one so that we only need to make a single new endpoint, otherwise we'll need _two_ new create auction endpoints to support the progressive changes in UI code. Take a look at his branch to see what he is doing. I think given you're both touching the same code, you should hand this code off to him, have him merge it and adjust his code and then move on to the next thing. Other than the fact that you have broken backwards compatibility, which is easily remedied, it is a working PR I believe!

ysavchenko:
Thanks for the comments, it makes sense to merge this change with the one @VadimGrozinok is working on.

ysavchenko:
Closing it in favour of PR 159

PR #173, opened 2021-07-28
feature: success tests for token-metadata

bezrazli4n0:
wip


ysavchenko:
I think we can review and merge the tests as they are now (just positive cases) and expand them in the future PRs. @bhgames  what do you think?

bhgames:
Wait...what are tests?

bhgames:
THis is great. Would love to see some docs on how to run thesee things because I have no idea :)

PR #190, opened 2021-08-05
Don't display store_owner_address on non-localhost sites.

mplumador:
This is to prevent people who are using their personal wallet address as the store owner address from accidentally leaking their wallet on their production sites.


bhgames:
I like being able to see that because then it allows me to easily simulate their sites for testing and debug (which is needed from time to time). Also this information is pretty easily accessible on the blockchain already - all I need to do is look up the create store transaction on the explorer (easy to find just by looking up the store id). 

If you don't want your personal wallet leaked, don't use your personal wallet as the administrative wallet for your store. Use a second wallet - which I'd recommend anyway, as you should never use your personal wallet to control a piece of software!

PR #194, opened 2021-08-06
Refactor/meta

exromany:
This reduced the time of loading and processing data from 8.5 to 7.4 seconds (on my machine)
All the magic in the last commit


js/packages/web/src/hooks/useArt.ts
```diff
@@ -136,7 +137,7 @@ export const useArt = (id?: PublicKey | string) => {
   const { metadata, editions, masterEditions, whitelistedCreatorsByCreator } =
     useMeta();
 
-  const key = typeof id === 'string' ? id : id?.toBase58() || '';
+  const key = pubkeyToString(id);
```
bhgames, line 140: We were just planning on replacing all the toStrings with .equals, which compares arrays and is probably cheaper. Why not do that?

bhgames:
Dont have time to look at this today but will try to take a good look at it tomorrow. I have a huge bunch of changes on my branch too that are going to get merged soon so it will likely cause conflicts, we would need to work out.

bhgames:
Alright I tested it out and it worked, but you now have a pretty big merge conflict which is that you nicely migrated everything out of meta.tsx but meta.tsx has some changes in it that need to be handled - a few new state entries, some new models (like AuctionManagerV2, SafetyDepositConfig, and BidRedemptionTicketV2), etc. Let me know when you've updated, and I'll give it another kick. Other than that, it looks good.

exromany:
the conflicts are fixed

PR #195, opened 2021-08-07
Making Metaplex packages publishable to crates.io

zomglings:
The goal of this pull request is to make it possible for anyone to interface with the Metaplex programs (`metaplex`, `token-metadata`, `token-vault`, `auction`) by import packages from crates.io into Rust programs.

### DIFF
--- a/rust/metaplex/program/Cargo.toml
+++ b/rust/metaplex/program/Cargo.toml
@@ -13,14 +13,14 @@ no-entrypoint = []
 test-bpf = []
 
 [dependencies]
-spl-auction = { path = "../../auction/program", features = [ "no-entrypoint" ] }
+spl-auction = { path = "../../auction/program", features = [ "no-entrypoint" ], version = "0.0.1" }
 num-derive = "0.3"
 num-traits = "0.2"
 arrayref = "0.3.6"
 solana-program = "1.6.10"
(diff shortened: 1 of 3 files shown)


zomglings:
As a first step, I am making it so that `cargo package` works in each of the individual packages in the workspace.

Since there are many different packages in the workspace, `cargo package` is still not working for packages that depend on other packages in the workspace (e.g. `token-metadata` depends on `token-vault` and so although `cargo package` works for `token-vault`, it doesn't work for `token-metadata` because `token-vault` has not been published to crates.io).

This is an open issue on Cargo: https://github.com/rust-lang/cargo/issues/1169

It makes it difficult to test. `cargo package` does not respect the `[patch.crates-io]` section in `Cargo.toml`.

rust/metaplex/program/Cargo.toml
```diff
@@ -13,14 +13,14 @@ no-entrypoint = []
 test-bpf = []
 
 [dependencies]
-spl-auction = { path = "../../auction/program", features = [ "no-entrypoint" ] }
+spl-auction = { path = "../../auction/program", features = [ "no-entrypoint" ], version = "0.0.1" }
```
bhgames, line 16: So inside the repo we would still want to be pointing at relative so we can do easy code edits

PR #210, opened 2021-08-13
Add gamma output to renderer

sol-survivor:
Fixes dark rendering of models


sol-survivor:
This fix works when running but will not cleanly build.

PR #211, opened 2021-08-13
Add gamma output to webgl renderer

sol-survivor:
Fix dark model view bug


js/packages/web/src/components/MeshViewer/index.tsx
```diff
@@ -41,7 +41,7 @@ export class MeshViewer extends React.Component<MeshViewerProps, {}> {
 
   private gltfLoader: GLTFLoader = new GLTFLoader();
 
-  private renderer?: THREE.WebGLRenderer;
+  private renderer?: any;
```
bhgames, line 44: Why was this necessary?

PR #213, opened 2021-08-15
Improve init store

exromany:
There are 2 main points:

1) app can read STORE_ADDRESS from env variables - instead of recalculating on every load from STORE_OWNER_ADDRESS

PS: since STORE_ADDRESS depends only on owner address and METAPLEX_ID - we can store it permanently

_correct me if I'm wrong_ 

2) `Init store` process can be started even  before storing owner address in env

If there is no store configuration (or store doesn't exists in blockchain), then the store will be created from the current wallet address

after initialization, the configuration text will be shown, which must be saved to a .env file

Screenshot 2021-08-15 at 17 31 07


exromany:
@bhgames @bartosz-lipinski can you look at this?

bhgames:
Alright sorry will take a look..been a long week

PR #214, opened 2021-08-15
Patch Users are able to view and auction owned NFTs

kespinola:
## Issue

With the latest release of Metaplex client, users are no longer able to see their owned NFT or put them up for auction.

## Resolution
- Setting `REACT_APP_BIG_STORE` to true gets owned NFTs visible in artworks tab.
- The select users account callback was returning empty when scanning the available accounts. The account.pubkey matches the current wallet so checking on that vs the info.owner (which always results in empty filter)

## On-Going Investigation

The select artwork component relies on `useUserArts` to populate the pick list of nfts for auction. Currently, the `metadata` returned by `useMetada` has all 4 of my owned NFTs but are filtered out before returning to the select artwork component. The first issue solved was getting userAccount returning an account with the change to account filtering but still not passing the other filter checks. 

`` accountByMint.get(m.info.mint)` is always undefined so no metadata are being pushed to the safety deposit box list which stores the NFT to be selected.

https://github.com/metaplex-foundation/metaplex/blob/81023eb3e52c31b605e1dcf2eb1e7425153600cd/js/packages/web/src/hooks/useUserArts.ts#L44


kespinola:
image


Example of accountByMint with the change to looking by pubkey. Initial metadata with my 4 owned NFTs then empty result after filtering to ownedMetadta.

The `has` operation always returns false which results in all metadata being filtered out.

https://github.com/metaplex-foundation/metaplex/blob/81023eb3e52c31b605e1dcf2eb1e7425153600cd/js/packages/web/src/hooks/useUserArts.ts#L29

PR #220, opened 2021-08-18
Model viewer

Arrowana:
Sorry, after more investigation and usage, it seems like the viewer used is a pain.

At the end we should go for model-viewer, supported by most browsers, also used by Phantom + Solflare X.
It seems to be the way to go to have minimal code and end up with consistent rendering everywhere.

https://modelviewer.dev/

Also reverting #120 

to test set `REACT_APP_STORE_OWNER_ADDRESS_ADDRESS=Hmvy3zMP9kwGWbD8Y8rpEuZW4Dt2j32xBSqzosNjoGgz`
So you can test looking at https://nft.ram.so/ assets


bhgames:
This looks good, just resolve conflicts!

bartosz-lipinski (approved):

Arrowana:
> This looks good, just resolve conflicts!

Done

PR #270, opened 2021-09-01
minor corrections (tx['txid'] and remove duplicate @solana/web3.js import)

nftintern:
* in the arweave transaction, we need `tx['txid']` instead of `tx` or it will error
* in cli's package.json we need to remove the duplicate `@solana/web3.js` import or it will fail with a name collision

### DIFF
--- a/js/packages/cli/package.json
+++ b/js/packages/cli/package.json
@@ -21,7 +21,6 @@
   },
   "dependencies": {
     "@project-serum/anchor": "^0.13.2",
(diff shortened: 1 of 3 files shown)


BoitToken:
> * in the arweave transaction, we need `tx['txid']` instead of `tx` or it will error
> * in cli's package.json we need to remove the duplicate `@solana/web3.js` import or it will fail with a name collision

After making these changes in program. It is running but while going on a website after the integration of github and vs code, blank screen is coming. Please help with this

PR #302, opened 2021-09-04
[Fix] Ensure account before checking owner

kespinola:
# Changes
Ensure there is an account before checking the owner. Currently, some accounts being processed are null which throw an error and prevent the auctions from loading.


kespinola:
Addressed in another pull request by using helper function to check for null.

PR #333, opened 2021-09-07
Fix rounding issue in cli

21e8:


js/packages/cli/src/cli.ts
```diff
@@ -552,7 +552,7 @@ program
     const { keypair } = cmd.opts();
     const solPriceStr = cmd.getOptionValue('price') || '1';
 
-    const lamports = Math.ceil(parseFloat(solPriceStr) * LAMPORTS_PER_SOL);
+    const lamports = +solPriceStr * LAMPORTS_PER_SOL;
```
BertrandBev, line 555: What's that `+` for? Don't you want `lamports` to be an integer? Not sure why the rounding to the next lamport would be a problem here

BertrandBev (changes requested):

js/packages/cli/src/cli.ts
21e8: the `+` is casting to number.
Please read the code correctly this price is for a SOL price, e.g. 1.5 SOL, which is why we multiply with `LAMPORTS_PER_INTEGER`

BertrandBev: I think the `+` implicit parsing is rather unclear (though I like the trick from a code golfing perspective !). Besides, we still don't necessarily end up with an integer number (if price is "1e-10" say).

brandonculver: I don't think this is a problem with the new code. I set the price to 0.1 and it worked as expected.

PR #334, opened 2021-09-07
Feat/metaplex cli overhaul merged

bhgames:


js/packages/cli/src/helpers/various.ts
```diff
@@ -0,0 +1,93 @@
...
+    ? JSON.parse(fs.readFileSync(path).toString())
+    : undefined;
+}
+
+export function saveCache(cacheName: string, env: string, cacheContent) {
+  fs.writeFileSync(cachePath(env, cacheName), JSON.stringify(cacheContent));
+}
+
+export function parsePrice(price): number {
+  return Math.ceil(parseFloat(price) * LAMPORTS_PER_SOL);
```
uaktags, line 78: why are we still rounding this to an integer?

PR #354, opened 2021-09-09
initial commit -not tested

thedonmon:
Need to run some testing, going to see how it handles uploading files


js/packages/cli/src/cli.ts
```diff
@@ -129,9 +130,7 @@ program
     let config = cacheContent.program.config
       ? new PublicKey(cacheContent.program.config)
       : undefined;
-
-    for (let i = 0; i < SIZE; i++) {
-      const image = images[i];
+    await asyncForEach(images, async (image, i) => {
```
bartosz-lipinski, line 133: This should be chunked to avoid spamming the server

PR #360, opened 2021-09-10
Metaplex Nft-packs

VadimGrozinok:
This PR adds new nft-packs program to Metaplex system.
Detailed description you can find in README: https://github.com/atticlab/metaplex/tree/feature/nft-packs/rust/nft-packs


VMey:
@VadimGrozinok 
My team has a conceptual approach we're targeting that would allow the following: breaking one NFT into smaller elements, trading the elemental pieces, and then recombining them into a new NFT.

We wrote a quick article about the concept here: https://link.medium.com/fZ810XgRpjb 

Jordan told me about this project via Twitter DMs. We'd really like to build something that would be reusable by other projects, and it seems you're pretty far along.

It is really critical to preserve rarity.  Here's how we see the mechanism working at the most basic level:
- When dismantling, the NFT being dismantled is burned and new NFTs are minted.
- When user goes to rebuild, we first check to see if what they're trying to build already exists.  We won't let anyone build something that someone already has.
- Upon rebuilding, the elements being merged are burned and a new full NFT is minted.

How compatible is this idea with your goals here?

cc @arbonserg @vitaliygor

PR #337, opened 2021-09-08
another overhaul

WrRaThY:
- added backend signing for a single metadata id
- some small refactoring
- extracted upload from the CLI (maybe will do the same with other methods)
- automatic re-uploading on failures

PR #365, opened 2021-09-11
Candy machine post sale mass signer

solserer-labs:
This pull request implements a script that will:

- Get accounts by candy machine creator address
- Get only verified ones
- Get only unverified ones with creator address (keypair address)
- Grab n at a time and batch sign and send transaction

Note that it also adds another build mode `package:linuxb` as building with the --no-bytecode flag didn't work for me. Idk what's the best way to about this, any suggestions?

### DIFF
--- a/js/packages/cli/src/cli.ts
+++ b/js/packages/cli/src/cli.ts
@@ -1,224 +1,49 @@
 #!/usr/bin/env node
 import * as fs from 'fs';
 import * as path from 'path';
-import fetch from 'node-fetch';
 import FormData from 'form-data';
 import { program } from 'commander';
 import * as anchor from '@project-serum/anchor';
 import BN from 'bn.js';
 import { MintLayout, Token } from '@solana/spl-token';
 
-import { sendTransactionWithRetryWithKeypair, fromUTF8Array } from './helper';
 import {
-  LAMPORTS_PER_SOL,
-  PublicKey,
-  SystemProgram,
-  SYSVAR_RENT_PUBKEY,
-  TransactionInstruction,
-} from '@solana/web3.js';
+  chunks,
+  fromUTF8Array,
+  loadCache,
+  parsePrice,
+  saveCache,
+  upload,
+} from './helpers/various';
+import { Keypair, PublicKey, SystemProgram } from '@solana/web3.js';
+import { createAssociatedTokenAccountInstruction } from './helpers/instructions';
+import {
+  CACHE_PATH,
+  CONFIG_ARRAY_START,
+  CONFIG_LINE_SIZE,
+  EXTENSION_JSON,
+  EXTENSION_PNG,
+  PAYMENT_WALLET,
+  TOKEN_METADATA_PROGRAM_ID,
+  TOKEN_PROGRAM_ID,
+} from './helpers/constants';
+import { sendTransactionWithRetryWithKeypair } from './helpers/transactions';
+import {
+  createConfig,
+  getCandyMachineAddress,
+  getMasterEdition,
+  getMetadata,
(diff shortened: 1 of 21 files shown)


js/packages/cli/src/cli.ts
```diff
@@ -583,4 +583,42 @@ program
...
+    '-e, --env <string>',
+    'Solana cluster env name',
+    'devnet', //mainnet-beta, testnet, devnet
+  )
+  .option(
+    '-k, --keypair <path>',
+    `Solana wallet location`,
+    '',
+  )
+  .option('-cndy, --candy-address <string>', 'Candy machine address', '')
```
WrRaThY, line 598: address may be taken from cache

js/packages/cli/src/cli.ts
```diff
@@ -583,4 +583,42 @@ program
...
+      console.log("Batch size needs to be an integer!")
+      return;
+    }
+    const walletKeyPair = loadWalletKey(keypair);
+    const anchorProgram = await loadAnchorProgram(walletKeyPair, env);
+    console.log("Creator pubkey: ", walletKeyPair.publicKey.toBase58())
+    console.log("Environment: ", env)
+    console.log("Candy machine address: ", candyAddress)
+    console.log("Batch Size: ", bVal)
+    await signAllMetadataFromCandyMachine(anchorProgram.provider.connection, walletKeyPair, candyAddress, bVal)
```
WrRaThY, line 621: providing anchorProgram would be cleaner
rename bVal batchSizeInt or batchSizeParsed or smth like that

js/packages/cli/src/signer.ts
```diff
@@ -0,0 +1,435 @@
...
+ Get only unverified ones with creator address
+ Grab n at a time and batch sign and send transaction
+
+ PS: Don't sign candy machine addresses that you do not know about. Signing verifies your participation.
+*/
+const MAX_NAME_LENGTH = 32;
+const MAX_URI_LENGTH = 200;
+const MAX_SYMBOL_LENGTH = 10;
+const MAX_CREATOR_LEN = 32 + 1 + 1;
+const METADATA_PROGRAM_ID = "metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s"
```
WrRaThY, line 17: move them to constants (METADATA_PROGRAM_ID  is already there)

js/packages/cli/src/signer.ts
```diff
@@ -0,0 +1,435 @@
...
+}
+
+class Creator {
+  address: string;
+  verified: boolean;
+  share: number;
+
+  constructor(args: {
+    address: string;
+    verified: boolean;
```
WrRaThY, line 103: all of those should go to `types.ts`

WrRaThY (commented):
a couple of things that could be done better, but you might just want to merge your changes here:
https://github.com/metaplex-foundation/metaplex/pull/337
I tried to tackle that before just couldn't figure out how to find all those account (which you did), I just think that the structure I proposed those is a bit clearer

js/packages/cli/src/cli.ts
solserer-labs: afaik address isn't being added to the .cache file only the config address.

i added a json file that will store all the candy machines you make from the cli inside the .cache folder

js/packages/cli/src/cli.ts
solserer-labs: i've only now started to look into anchor

solserer-labs:
ok merged with latest master and tested it seems to work

solserer-labs:
moved the signer to commands and using cache for candy machine address now

PR #369, opened 2021-09-12
remove `use-wallet` eth package

21e8:
This package is an eth package and we use the provider but nothing else.
I can imagine that this might be connecting to some ETH rpc, causing lag.

https://www.npmjs.com/package/use-wallet

I could not find any difference in the behvaiour of the app, I think it's safe to delete since we have our own wallet providers here.


bartosz-lipinski:
Good catch, thank you.

PR #375, opened 2021-09-12
Update the --no-bytecode to -d

hamzasaleem2:


bhgames:
I think this was fixed with as different PR

PR #390, opened 2021-09-13
Fix lint

jordansexton:
No change to functionality here -- just running `yarn lint:fix`.


jordansexton:
Closed since this would affect other open PRs. CLI can be linted later.

PR #391, opened 2021-09-13
Update wallet-adapter dependencies and interface

jordansexton:
I merged https://github.com/metaplex-foundation/metaplex/pull/390 since it contains many of the same lint fixes. If/when that is merged to master, the changeset here will be much smaller.


jordansexton:
Merged master and updated wallet-adapter deps again.

PR #395, opened 2021-09-14
Added HTML file support to the animation_url parameter

stegaBOB:
This PR adds support for minting NFTs with HTML files in the URL parameter or by adding the files directly (and then upload to Arweave), along with viewing them. It uses iframes that are sandboxed to prevent any potential malicious code from running. This would allow OpenSea creators that have utilized the animation_url parameter for HTML files to more easily migrate to Solana.


stegaBOB:
https://user-images.githubusercontent.com/41593264/133282223-52fcadd5-a0e5-46d6-aee6-409b2f4c3f6c.mp4

PR #397, opened 2021-09-14
Command to verify CandyMachine price

grimAgent:
Heard @bhgames mention during an Office Hours that he would like to see this implemented, so here we are.

Also added an optional parameter for the cache-path so this can be run from anywhere and not just the folder where the assets are. I found that helpful during my tests, hopefully you will too!


grimAgent:
https://user-images.githubusercontent.com/90475821/133196681-aa621115-a239-44ad-bd8b-db0734746f95.mov

PR #398, opened 2021-09-14
Candy Machine Authority as a verified creator

devcresix:
This issue popped up when two candy machines were created from a single account. By implementing this, the creator can verify his own account if it is in the creators array. Then all the NFTs generated with the authority of the same Address will be verified.


bhgames:
I'm pretty sure this would explode because in this transaction the authority of the candy machine would only be able to sign for this and set verified = true if you were calling mint_one_token as that authority and signing fort his very transaction. Some other customer could not do this.

Issue #382, opened 2021-09-13
Feature request: URI and name templates


soualid:
Hello,

For large collections of collectible (10K or more), instead of storing each single configuration on-chain, it would be great to have a way to set the URI and names as template using the index of the token in the candy machine.

A possible implementation could be a `is_templated_metadata` in `ConfigData` that triggers a  computed version of `get_config_line` instead of the one stored on chain.

ie, `is_templated_metadata = true` and this single configuration line is stored in the configuration array:

```
{
  "name": "My dope NFT #{i}", 
  "uri: "https://some-server.com/metadatas/{i}.json"
}
```

This single configuration line could then be used in `get_config_line` to derive the proper configuration line from the template using the `index` parameter.

Implementing this would save a lot of storage for a typical use case in the NFT world, which is large collections of collectible. Actually, the NFT candy machine is barely usable in these use cases because storing 10K metadatas on chain would cost ~2000USD.

Thanks for reading!

soualid:
I implemented this approach locally and tested it on devnet (ie. sample [token #4](https://explorer.solana.com/address/8JndBAPe8E91ACVcafzeKcPFBXtq1KQtXBFJCGGZgh2t?cluster=devnet) and sample [token #20](https://explorer.solana.com/address/BVYCTbDJWoa5eAtmeJwmUWRuWQUTVmQWUSbND5qg2N2S?cluster=devnet)), I'll try to clean up the rust implementation a bit, and finalize the implementation of the required options on the CLI tool before submitting a PR for early feedbacks.

samuelvanderwaal:
The storage cost concern should be solved by the [withdraw function](https://docs.metaplex.com/create-candy/withdraw). The PR was closed due to inactivity so I am closing this issue as well.

PR #399, opened 2021-09-14
Implemented a templated metadatas feature for the candymachine on-chain program operatable from CLI

soualid:
I'm kinda new to rust and the whole Solana ecosystem, don't hesitate to tell me if something needs to be changed.

Implemented a templated metadatas feature for the candymachine on-chain program operatable from CLI. See https://github.com/metaplex-foundation/metaplex/issues/382.

# How to test or use this feature

Considering the following `metadatas.json` file:

```
{
  "name": "Test #{i}",
  "symbol": "TST",
  "properties": {
    "creators": [
      { "address": "{redacted}", "share": 100 }
    ]
  }
}
```

```
ts-node cli initialize_templated_metadata metadatas.json -u 'https://test/{i}.json' -k {redacted}
```

Will output: 

```
starting templated metadatas configuration generation
initializing new config
wallet public key: {redacted}
initialized config for a candy machine with publickey: {redacted}
Done.
```

Then: 

```
ts-node cli create_candy_machine -p 0.00006 -n 20 -k {redacted}
```

Will create the "templated" candy machine:

```
wallet public key: {redacted}
create_candy_machine finished. candy machine pubkey: {redacted}
```

That we can then use to create token that will use metadatas generated by the defined template:

```
for i in {1..20}; do ts-node cli mint_one_token -k {redacted}; done
```

We can use tests from `token-metadata` on some of the created tokens to check the generation process: 

```
cargo run --bin spl-token-metadata-test-client show --mint 2KbeNBa2RT3R3BAN9SafoSHapsVv8aVfQb3WhUZs5UHq --keypair {redacted}
cargo run --bin spl-token-metadata-test-client show --mint 2KEMDZqkzZqYDDmmNdPXuWN4DyrPKHTnb5UUbSLifQ3P --keypair {redacted}
```

Will output: 

```
...
name: "Test #4\u{0}\u{0}\u{0}...",
uri: "https://test/4.json\u{0}\u{0}\u{0}...",
...
name: "Test #7\u{0}\u{0}...",
uri: "https://test/7.json\u{0}\u{0}...",
...
```

A candy machine implementing this feature as been deployed here for testing purpose on devnet: `3gvU2gogD8LoYcuSH3rSDbfCqrW6UeT78o4YX822TrY4`.

Thanks for reading!

### DIFF
--- a/js/packages/cli/src/helpers/constants.ts
+++ b/js/packages/cli/src/helpers/constants.ts
@@ -21,7 +21,8 @@ export const CONFIG_ARRAY_START =
   8 + //max supply
   1 + //is mutable
   1 + // retain authority
-  4; // max number of lines;
+  4 + // max number of lines
+  1; // is templated
 export const CONFIG_LINE_SIZE = 4 + 32 + 4 + 200;
 
 export const CACHE_PATH = './.cache';
--- a/rust/nft-candy-machine/src/lib.rs
+++ b/rust/nft-candy-machine/src/lib.rs
@@ -95,6 +95,7 @@ pub mod nft_candy_machine {
 
         let config_line = get_config_line(
             &config.to_account_info(),
+            config.data.is_templated_metadata,
             candy_machine.items_redeemed as usize,
         )?;
 
@@ -397,11 +398,11 @@ pub mod nft_candy_machine {
 
         if get_config_count(&ctx.accounts.config.to_account_info().data.borrow())?
             < candy_machine.data.items_available as usize
+            && ctx.accounts.config.data.is_templated_metadata != true
         {
             return Err(ErrorCode::ConfigLineMismatch.into());
         }
-
-        let _config_line = match get_config_line(&ctx.accounts.config.to_account_info(), 0) {
+        let _config_line = match get_config_line(&ctx.accounts.config.to_account_info(), ctx.accounts.config.data.is_templated_metadata, 0) {
             Ok(val) => val,
             Err(_) => return Err(ErrorCode::ConfigMustHaveAtleastOneEntry.into()),
         };
@@ -526,7 +527,8 @@ pub const CONFIG_ARRAY_START: usize = 32 + // authority
 8 + //max supply
 1 + // is mutable
 1 + // retain authority
-4; // max number of lines;
+4 + // max number of lines
+1; // is templated;
 
 #[account]
 #[derive(Default)]
@@ -551,6 +553,7 @@ pub struct ConfigData {
     pub is_mutable: bool,
     pub retain_authority: bool,
     pub max_number_of_lines: u32,
+    pub is_templated_metadata: bool,
 }
 
 pub fn get_config_count(data: &Ref<&mut [u8]>) -> core::result::Result<usize, ProgramError> {
@@ -559,8 +562,16 @@ pub fn get_config_count(data: &Ref<&mut [u8]>) -> core::result::Result<usize, Pr
 
 pub fn get_config_line(
     a: &AccountInfo,
-    index: usize,
+    is_templated_metadatas:bool,
+    in_index: usize,
 ) -> core::result::Result<ConfigLine, ProgramError> {
+
+    let requested_index:usize = in_index + 1;
+    let index:usize = match is_templated_metadatas {
+        true => 0,
+        false => in_index
+    };
+
     let arr = a.data.borrow();
 
     let total = get_config_count(&arr)?;
@@ -571,8 +582,14 @@ pub fn get_config_line(
         ..CONFIG_ARRAY_START + 4 + (index + 1) * (CONFIG_LINE_SIZE)];
 
     let config_line: ConfigLine = ConfigLine::try_from_slice(data_array)?;
-
-    Ok(config_line)
+    if is_templated_metadatas {
+        Ok(ConfigLine {
+            name: config_line.name.replace("{i}", requested_index.to_string().as_str()).to_string(),
+            uri: config_line.uri.replace("{i}", requested_index.to_string().as_str()).to_string(),
+        })
+    } else {
+        Ok(config_line)
+    }
 }
 
 pub const CONFIG_LINE_SIZE: usize = 4 + MAX_NAME_LENGTH + 4 + MAX_URI_LENGTH;
--- a/js/packages/cli/src/cli.ts
+++ b/js/packages/cli/src/cli.ts
@@ -12,6 +12,7 @@ import { CACHE_PATH, CONFIG_ARRAY_START, CONFIG_LINE_SIZE, EXTENSION_JSON, EXTEN
 import { getCandyMachineAddress, loadAnchorProgram, loadWalletKey, } from './helpers/accounts';
 import { Config } from './types';
 import { upload } from './commands/upload';
+import { initializeTemplatedMetadataConfiguration } from './commands/template';
 import { loadCache, saveCache } from './helpers/cache';
 import { mint } from "./commands/mint";
 import { signAllUnapprovedMetadata, signMetadata } from "./commands/sign";
@@ -24,6 +25,27 @@ if (!fs.existsSync(CACHE_PATH)) {
 }
 
 log.setLevel(log.levels.INFO);
+
+programCommand('initialize_templated_metadata')
+  .argument(
+    '<metadata-template.json>',
+    'A file containing the metadata template, using {i} as an index placeholder',
+  )
+  .option(
+    '-u, --uri-template <string>',
+    'URI template for metadatas, using {i} as an index placeholder, ie: https://some_server/ipfs/metadatas/{i}.json',
+  )
+  .action(async (templatePath, _, cmd) => {
+    console.log('starting templated metadatas configuration generation');
+    const { keypair, env, uriTemplate, cacheName } = cmd.opts();
+    initializeTemplatedMetadataConfiguration(
+      templatePath,
+      uriTemplate,
+      cacheName,
+      env,
+      keypair,
+    );
+  });
 
 programCommand('upload')
   .argument(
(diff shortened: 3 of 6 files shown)


js/packages/cli/src/helpers/constants.ts
```diff
@@ -21,7 +21,8 @@ export const CONFIG_ARRAY_START =
   8 + //max supply
   1 + //is mutable
   1 + // retain authority
-  4; // max number of lines;
+  4 + // max number of lines
+  1; // is templated
```
bhgames, line 25: What you have not considered is this will break every candy machine in existence that isn't your own new kind. You need to think of a way to gracefully handle them.

bhgames: Only way I can see this working is with a ConfigV2.

bhgames (changes requested):

js/packages/cli/src/helpers/constants.ts
soualid: Many thanks for your feeback, and damn, you are right, sorry for that!

I'm a bit new to both solana and rust which may explain why I'm not sure what would be the best way to handle backward compatibility in this case. Is the `ConfigV2` you pointed a `ctx.accounts.configv2` dedicated account, that we could check to know if we're dealing with a configuration including the `is_templated` field?

Another way of implementing that could be to set the `max_number_of_lines` to `u32::MAX` when declaring a templated configuration, that value triggering the required mechanisms. It feels a bit hacky but could be less intrusive than a dedicated account.

Anyway, I would be more than happy to implement the preferred solution if this template mechanism is welcome in the basecode of metaplex.

bhgames: Well Configv2 would literally be a new type so it would have a different account type. This would mean, as far as anchor is concerned, new endpoints (you would duplicate them and instead of having Config you would use ConfigV2, and then update the CLI). This would mean that people using the older cli and endpoints would ocntinue experiencing normal operations and those using the newer CLI would exprience no issues either..

soualid: Got it, I actually implemented it using the latter way (a custom `max_number_of_lines` value triggering the required mechanisms), it is working but it was still needing cleanup and like I was expecting, it feels hacky.

I see here a good opportunity to introduce versioning of the candy machine configuration structure, I should have some time to implement it this way this week-end. Many thanks for your insights!

soualid: I have been able to implement backward compatbility using an `initialize_config_v2` method, and because duplicating all anchor endpoints felt a bit bloaty, I used the part of the configuration that is manually serialized to add the required information at the very end of the configuration,  this information is checked in a backward compatible way in `is_config_templated()` [here](https://github.com/soualid/metaplex/blob/a9a5e75d6ecc02636db1bb1bdcd9fca69a2de326/rust/nft-candy-machine/src/lib.rs#L567), which check this value, or return `false` if the size of the configuration indicate that we are dealing with a configuration created before the implementation of this feature.

soualid:
As detailed in this discussion https://github.com/metaplex-foundation/metaplex/pull/399#discussion_r711633919, the changes required by backward compatibility have been implemented in a new commit.

I have tested the different use cases (new program handling old and new configurations, as well as "template based" vs "standard" configurations) and everything seems to behave correctly. I'll do some more tests by monday to be sure.

Also, maybe I missed it, but I think a documentation for the CLI could be useful, english is not my mother tongue but I may write one describing the different use cases covered by the CLI if you want to.

Thanks!

js/packages/cli/src/helpers/constants.ts
bhgames: Ok i will take a look at this

js/packages/cli/src/helpers/instructions.ts
```diff
@@ -74,8 +74,8 @@ export async function createConfigAccount(
     4 +
     configData.maxNumberOfLines.toNumber() * CONFIG_LINE_SIZE +
     4 +
-    Math.ceil(configData.maxNumberOfLines.toNumber() / 8);
-
+    Math.ceil(configData.maxNumberOfLines.toNumber() / 8) +
+    1;
   return anchor.web3.SystemProgram.createAccount({
```
bhgames, line 79: Nit: Given the boolean comes earlier in the struct, this 1 should be in the place where it actually takes place

rust/nft-candy-machine/src/lib.rs
```diff
@@ -273,7 +279,11 @@ pub mod nft_candy_machine {
         for i in 0..4 {
             data[vec_start + i] = as_bytes[i]
         }
-
+        // previous CLI versions used a smaller configuration account, that's why we must check
+        // the configuration account size before writing this information
+        if data.len() >= vec_start + 5 {
```
bhgames, line 284: So I dislike this design pattern because it violates the borsh specification. The reason the number of lines u32 is right before the actual array is because that's how borsh stores vecs. Plopping this in between, while convenient, breaks things. What I would do instead is check if the date len is longer before you set the max number of lines, and if it is, offset the max number of lines by 1 byte and set it, so you can place the new boolean in that extra byte area.

bhgames (changes requested):
Needs a bit of rethink on storage choice. Ping me again when you've figured that out.

rust/nft-candy-machine/src/lib.rs
soualid: Sure, I'll change that asap, thanks for the feedback!

js/packages/cli/src/helpers/instructions.ts
soualid: We should be good on this.

rust/nft-candy-machine/src/lib.rs
soualid: Just pushed a commit that changes the position of this boolean, it should be OK even if I have to admit I'm not sure to have fully understanded the role of this mask.

Also rebased the branch on `master`. Thanks!

bhgames:
@soualid ping me when it's ready for re-review :)

soualid:
Hey @bhgames sorry! I think it is, or it should be! 🙂

I rebased my branch on `master` yesterday for it to be mergeable without conflict, and the boolean we added should now be stored at the proper place in the account data, and in a backward compatible way. 

Don't hesitate to tell me if it's not, I'll implement the required changes with pleasure. 

Many thanks for the reviews!

rust/nft-candy-machine/src/lib.rs
```diff
@@ -286,7 +300,8 @@ pub mod nft_candy_machine {
         let account = config.to_account_info();
         let current_count = get_config_count(&account.data.borrow())?;
         let mut data = account.data.borrow_mut();
-
+        let is_new_configuration = data.len() >= CONFIG_ARRAY_START + 4 + (current_count as usize) * CONFIG_LINE_SIZE + 5;
+        msg!("is_new_configuration: {} for {} and {} lines", is_new_configuration, data.len(), current_count);
```
bhgames, line 304: Yeah the more I look at this, the more I dislike it. I talked with @armaniferrante about how to do this the right way and he came up with this good example:

```rust
#[account]
pub struct MyAccountV1 {...}

#[account]
pub struct MyAccountV2 {...}

#[derive(Clone)]
pub enum MyAccount {
   V1(MyAccountV1),
   V2(MyAccountV2),
}

impl AccountDeserialize for MyAccount {
  fn try_deserialize(buf: &mut &[u8]) -> Result<Self, ProgramError> { 
    if discriminator(buf) == MyAccountV1::discriminator() {
       Ok(MyAccount::V1(MyAccountV1::try_deserialize(buf)))
    } else if discriminator(buf) == MyAccountV2::discriminator() {
       Ok(MyAccount::V2(MyAccountV2::try_deserialize(buf)))
    } else {
       Err(...)
     }
  }
  fn try_deserialize_unchecked(buf: &mut &[u8]) -> Result<Self, ProgramError> {
    // similar to above
  }
}

impl AccountSerialize for MyAccount {...}
```

Doing this would allow us to essentially turn the Config into an enum of ConfigV1 and ConfigV2. I think this is how it has to be done. It's more work but if we make more changes in the future it'll be much more extensible than this hacky way.

soualid: Cannot agree more, it's way more elegant, and less error prone, than what we ended up with. I'll try to implement it this way asap! 

Thanks for pointing out this solution.

soualid: I liked this idea, but unfortunately, if I understood everything well, I think it might hit some limitations in the anchor library that may need to be worked around.

First, we cannot use [tuple enum variants](https://github.com/project-serum/anchor/blob/90df0b19766ef2a0937312a3864f44e49bb16cad/ts/src/coder/idl.ts#L131) for the `Config` enum for now, but it can still be serialized properly by anchor if we define named variants, which ends with the following in the IDL:

```
{
  "name": "Config",
  "type": {
    "kind": "enum",
    "variants": [
      {
        "name": "V1",
        "fields": [
          {
            "name": "config",
            "type": {
              "defined": "ConfigV1"
            }
          }
        ]
      },
      {
        "name": "V2",
        "fields": [
          {
            "name": "config",
            "type": {
              "defined": "ConfigV2"
            }
          }
        ]
      }
    ]
  }
}
```

Unfortunately, it seems anchor is not serializing the related `ConfigV1` and `ConfigV2` types in the generated IDL, and the following error pops client-side when instantiating the program stub: 

```
IdlError: Type not found: {"name":"config","type":{"defined":"ConfigV1"}}
```

I have to check why in the anchor library and maybe ask them about that, but to make a long story short, I cannot guarantee that I will end up with a working implementation this way, but I'll update this PR if I do!

I kept both working implementations to spare branches in my fork, don't hesitate to ask me if you want me to create a PR for one of them. One of them used the `u32::MAX` (or `0`, which is also unused) values as trigger in the `max_number_of_lines` field of the configuration, and the other is this PR, which serializes this trigger as a boolean in the custom borsh data of the configuration in a backward compatible way.

Without considering this limitation, after trying different ways, my feeling is that the `max_number_of_lines` approach was less error prone than the custom borsh serialization, and that it is more compact and readable than a custom enumeration, even if it is less elegant and extensible. But this can definitely be the consequence of my rudimentary knowledge of the solana ecosystem.

In the between, people can still use a custom program using token metadatas, or a forked version of the candy machine, if they need to emit large collectible without storing every single configuration line on the blockchain.

It was interesting to work on this and to try these different approaches, thank you again for your insights on this! As I said, I'll update the PR if I find enough spare time to dig this serialization issue in anchor, and I'll keep an eye on this repo to see if this template feature is implemented somehow in the future!

Cheers!

bhgames:
@soualid, this good to review?

soualid:
@bhgames  not yet I'm struggling with this serialization issue on the anchor side! 

https://github.com/metaplex-foundation/metaplex/pull/399#discussion_r715390786
https://github.com/project-serum/anchor/issues/806

PR #407, opened 2021-09-15
Provide error message to user when upload of artwork to Arweave fails

kespinola:
### Changes

- Introduce rejections to the mint auction when the response from uploadFile endpoint is not ok or the internal error field is populated. I'm not sure on the error messages the uploadFile endpoint provides.
- Replace the progress bar with a spinner. Holaplex has received many frustrating reports around 99% completion so I think switching to a spinner until the progress of the nft creation can truly be tracked is a better user experience.
- If the minting auction throws any errors the user will see the error message instead of congratulations. 

Screenshot from 2021-09-15 08-35-34
Screenshot from 2021-09-15 08-31-54

### Questions

What are the implications of a partially minted NFT? The calls to arweave upload endpoint happen after the initial mint call is made. If the upload errors then the remainder of the function will not execute. If tested partial mint by forcing the rejection and so no ghost nft in my wallet.

https://github.com/metaplex-foundation/metaplex/blob/master/js/packages/web/src/actions/nft.tsx#L143


bhgames:
@kespinola send me your wallet for a medal

PR #434, opened 2021-09-19
Replace uploadFile2 with 4

adamjeffries:


js/packages/web/src/actions/nft.tsx
```diff
@@ -323,7 +326,7 @@ export const prepPayForFilesTxn = async (
       SystemProgram.transfer({
         fromPubkey: wallet.publicKey,
         toPubkey: AR_SOL_HOLDER_ID,
-        lamports: await getAssetCostToStore(files),
```
bartosz-lipinski, line 326: I would still keep this, it was calculating valid cost

adamjeffries: In the CLI - storage cost appears to be fixed to "10" - is the calculation needed there and just didn't get around to it?

PR #441, opened 2021-09-19
Add `verify_token_metadata` command

jnwng:
created my first candy machine this weekend using `metaplex`! the CLI does a wonderful job of getting everything through to the finish line. i had a few trip-ups, though, particularly around the format of the metadata files expected by the candy machine (`0.json`, `1.json`, etc), as well as small errors that are somewhat simple to verify when you have a small amount of tokens but become much harder to verify when you get into the thousands.

you'll probably note that most of this is lifted from @levicook's [guide](https://hackmd.io/@levicook/HJcDneEWF) to the candy machine, which is generally accurate but just slightly out-of-date (who can blame them? things are moving fast). this script is intended to replace the [manual validation steps](https://hackmd.io/@levicook/HJcDneEWF#Validating-your-project-assets) that are laid out in the blog post, as well as various other slip-ups that i've noticed while watching the #development channel in Discord. 

longer-term, i'd like `metaplex upload` to run this command _first_, but that may be a breaking change in certain cases so i'm opting to keep it a separate invocation entirely until folks are pretty happy that it catches all the corner cases.

here's [the PR](https://github.com/metaplex-foundation/metaplex-docs/pull/9) for updating the JSON schema in the Metaplex docs, too.


js/packages/cli/src/commands/verifyTokenMetadata/index.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  verifyConsistentShares(collatedCreators);
+};
+
+export const verifyTokenMetadata = ({
+  files,
+  uploadElementsCount = null,
+}): Boolean => {
+  // Will we need to deal with the cache?
```
jnwng, line 159: i'm not sure whether we need to take into account the state of the cache or not. my sense is "no", but i could use a second pair of eyes

js/packages/cli/src/commands/verifyTokenMetadata/token-metadata.schema.json
```diff
@@ -0,0 +1,67 @@
+{
+  "title": "Token Metadata",
```
jnwng, line 2: this schema is non-exhaustive (doesn't capture things like `attributes`) but does validate against the critical properties in the config

.github/workflows/js-pull-request.yml
```diff
@@ -0,0 +1,29 @@
+name: Pull Request (JS)
+
+on:
+  pull_request:
+    paths:
+      - js/*
+  push:
+    branches: [master]
+    paths:
```
bhgames, line 9: what is this

bhgames (commented):
Fantastic work. All I will need here is a video showcasing it working and I will approve it, also will need your address to shoot you a medal.

.github/workflows/js-pull-request.yml
jnwng: this section is for running the tests in the `cli` package, but only when those files change (don't need to run these tests for rust-related changes, for example). this is not running on this PR because `pull_request` only works in the context of the source repository (`metaplex`) and until it reaches the default branch it won't run on forked PRs. if this lands, though, it'll make sure we're running tests whenever someone makes a change to the CLI. 

you can see this test running successfully [here](https://github.com/jnwng/metaplex/pull/1/checks?check_run_id=3708789793)

jnwng:
here's [a video](https://streamable.com/vekazl) that goes through:
- building the CLI
- packaging the CLI
- running the `verify_token_metadata` command against a local folder of token metadata broken in a few different ways (mismatched files, invalid aggregate creator share, `image.png` placeholders, inconsistencies _across_ tokens)
- running the `upload` command with validated token metadata

the video is super dry and low-res, i surpassed the video limit in github...  i suggest watching it at 2x if possible :)

just to re-iterate, the schema added here isn't completely exhaustive. since writing this PR there's a few other things i want to check (image size, other properties like attributes), but i think this is a great start for the sorts of errors i typically see in discord!

my address is `5dNGzQh9sonyFUcTHrH6wiCczokUMSc79miMEysyVYjK`

jnwng:
here's [a video](https://streamable.com/vekazl) that goes through:
- building the CLI
- packaging the CLI
- running the `verify_token_metadata` command against a local folder of token metadata broken in a few different ways (mismatched files, invalid aggregate creator share, `image.png` placeholders, inconsistencies _across_ tokens)
- running the `upload` command with validated token metadata

the video is super dry and low-res, i surpassed the video limit in github...  i suggest watching it at 2x if possible :)

just to re-iterate, the schema added here isn't completely exhaustive. since writing this PR there's a few other things i want to check (image size, other properties like attributes), but i think this is a great start for the sorts of errors i typically see in discord!

my address is `5dNGzQh9sonyFUcTHrH6wiCczokUMSc79miMEysyVYjK`

jnwng:
if there's a better site to upload videos i can re-upload, i'm all ears. im watching this on streamable and hate it

jnwng:
@bhgames this is ready to go! happy to make any adjustments if needed.

PR #453, opened 2021-09-21
Power Speed Run Based on The Store Owner

kespinola:
### Goal
All stores use speed run when the only NFTs and auctions the storefront should display are the storefront owners.

### Changes
- Assemble auction and metadata queries based on the store owner address.

### Tasks
- [X] Replace WHITELISTED_METATA with 5 filtered queries to get metadata from token metatdata program based on the creators field.
- [ ] Replace WHITELISTED_AUCTION_MANAGER with a filtered query to metaplex program based on the store address. [AuctionManagerV2 Struct](https://github.com/metaplex-foundation/metaplex/blob/08cce01bc8ed08ed683a997f2a1e16e2579fcaa9/rust/metaplex/program/src/state.rs#L196)
- [ ] Replace WHITELISTED_AUCTION after querying auction managers loop through results and fetch auction based on auction field set on auction manager.
- [ ] Replace  AUCTION_TO_METADATA  
- [ ] Replace AUCTION_TO_VAULT

### Screen Shots
image

### Questions
- Most storefronts only have 1 creator which is the store owner but storefronts are capable of displaying metadata for any whitelisted creator. However, whitelisted creators are a pda account whose address is derived from store addresss and wallet address. This requires comparing the address of all accounts in Metaplex and comparing it with the computed address for the whitelisted creator account. Is this assessment correct? Is there a means of using memcmp filters to find whitelisted creators based on the store address?
- I'm testing using the same owner address as https://bhgames.github.io/metaplex#/artworks. This version of the code is pulling up "Simple test" and not "Pit and Viper". What is off?
- Any of the Tasks with no strat for replacing our are still unknown. Ideas welcomed!

### DIFF
--- a/js/packages/common/src/contexts/meta/loadAccounts.ts
+++ b/js/packages/common/src/contexts/meta/loadAccounts.ts
@@ -36,7 +36,7 @@ import { ParsedAccount } from '../accounts/types';
 import { getEmptyMetaState } from './getEmptyMetaState';
 import { getMultipleAccounts } from '../accounts/getMultipleAccounts';
 
-export const USE_SPEED_RUN = false;
+export const USE_SPEED_RUN = true;
 const WHITELISTED_METADATA = ['98vYFjBYS9TguUMWQRPjy2SZuxKuUMcqR4vnQiLjZbte'];
 const WHITELISTED_AUCTION = ['D8wMB5iLZnsV7XQjpwqXaDynUtFuDs7cRXvEGNj1NF1e'];
 const AUCTION_TO_METADATA: Record<string, string[]> = {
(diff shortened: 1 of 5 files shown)


kespinola:
There is no means of querying extra auction information (eg safety deposit deposits, bidder redemptions, safety deposit box configs, bidder pots, bidder meta) without making n * m calls to blockchain where n is the number of auctions and m is the total auction extra account types.

What Im looking into now is doing an infinite scroll on the auctions list view which queries for all the auctions in load Account and then renders auctions in batches of 3-5 to get complete auctions (ie what is needed for AuctionViews).

PR #467, opened 2021-09-22
Separate file upload & candy machine file population

BertrandBev:
For now the file upload and the candy machine file population is done in the same function. Separating that process in two steps allows for a switch to a different storage medium


facundocajen:
so can this be done? i cant find a proper guide on how to use a private rpc server for hosting the images rather than going to arweave

PR #413, opened 2021-09-16
Updating metaplex CLI to support Infura IPFS + `retainAuthority` flag

edgarpavlovsky:
Summary:

* Extending metaplex CLI to support uploading to IPFS via Infura in addition to Arweave
* Making `retainAuthority` configurable in CLI instead of being hardcoded to `true`
* Updating cli version (looks like this hasn't been done in prior updates - should we keep it at `0.0.1`?)

h/t @21e8 for the IPFS [gist](https://gist.github.com/21e8/7b8c3c9f73dcb1b8a213c723768529d6) this week

Detail:

Hey folks. With arweave having image issues this past week, it seems a growing amount of us moved to using IPFS with candy machine. @21e8 modified the metaplex CLI to use IPFS via Infura in a [gist](https://gist.github.com/21e8/7b8c3c9f73dcb1b8a213c723768529d6) a few days ago & I thought I'd extend the CLI with similar functionality to make it easy for users to choose between using arweave or IPFS as their metadata storage.

While I was at it, I noticed a low-risk opportunity to add a `--no-retain-authority` flag and pass `eslint` checks.

Finally - I upgraded the cli version to `0.0.2`, but noticed it hasn't been upgraded in prior updates. lmk if we should keep at `0.0.1`.

This'll be my first PR to metaplex & I didn't see a contribution guide in the repo (sorry if I missed it), so lmk what I else I need to provide here.

PR #473, opened 2021-09-23
Fix resolution of duplicate react types

jnwng:
pulled from [here](https://stackoverflow.com/questions/52399839/duplicate-identifier-librarymanagedattributes), but restores the ability to successfully run `yarn build` after a fresh pull from yarn.

digging into the `yarn.lock`, looks like a new version of `@solana/wallet-adapter-material-ui` which has `@types/react` tied to [a specific version](https://github.com/solana-labs/wallet-adapter/blob/7eb8ec45a15e555fd9ace4ac48048380f7162b61/packages/ui/material-ui/package.json#L33), but in most cases this should be declared as `'@types/react': "*"` to let the consuming package (in this case, metaplex) choose which version. this is also [an issue](https://github.com/solana-labs/wallet-adapter/blob/7eb8ec45a15e555fd9ace4ac48048380f7162b61/packages/ui/ant-design/package.json#L33) with `@solana/wallet-adapter-ant-design` /cc @jordansexton 

in any case, setting the resolution here forces packages to refer to a single version of `@types/react`, which resolves the issue with typescript that comes up when you run `yarn build`.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -73,5 +73,8 @@
     "ts-jest": "^24.0.0",
     "ts-node": "^9.0.0",
     "typescript": "^4.1.3"
+  },
+  "resolutions": {
+    "@types/react": "17.0.21"
   }
 }
--- a/js/yarn.lock
+++ b/js/yarn.lock
@@ -4315,16 +4315,7 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react@*":
-  version "17.0.9"
-  resolved "https://registry.yarnpkg.com/@types/react/-/react-17.0.9.tgz#1147fb520024a62c9b3841f5cb4db89b73ddb87f"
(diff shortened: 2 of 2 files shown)


iammerrick:
Was able to get the CLI building again! Thanks!

jordansexton:
Good catch! I don't know about `*` but we should definitely make the dependency more liberal in wallet-adapter. I'd be happy to accept a PR there if you're interested

jnwng:
if https://github.com/solana-labs/wallet-adapter/pull/95 gets merged, we will still need to update `yarn.lock` to reconcile the `@types/react` package

bhgames:
can you rsolve conflcits?

jnwng:
we don't need this PR anymore, looks like the change needed here got merged as part of #413: https://github.com/metaplex-foundation/metaplex/blob/c81498932fa086d67630bc6a90c2416254338c1f/js/package.json#L32-L34

PR #474, opened 2021-09-23
fix subscription accounts not being processed

iamoperand:
All the processors `(processAuctions, processMetaData, processVaultData, processMetaplexAccounts)` want the owner to be in string format in order to verify if they belong to any.

This PR fixes that for accounts coming via `onProgramAccountChange`


bhgames:
send wallet for medal

iamoperand:
thanks! I'll dm you on twitter :)

PR #442, opened 2021-09-20
Feat/sign all

WrRaThY:
- renamed signing all NFTs to something nicer
- added daemon mode (it endlessly approves)
- minor refactoring (reduced some copy-pasting from the original sign_all implementation)
- ran `yarn lint:fix`... I hate JS linters...

PR #476, opened 2021-09-23
Feat/sign all v2

WrRaThY:
- refactored sign_all to be simpler and more readable
- added the daemon option, so you ran run it and forget and have everything signed
- added setting date to 'now' to update candy machine


WrRaThY:
just an update on this: https://github.com/metaplex-foundation/metaplex/pull/442

WrRaThY:
@bhgames fixed all merge conflicts and did a full run
at the end you see that when launched - the script found one currently existing NFT
then in a different terminal I minted another NFT, waited a couple of seconds and the script picked it up to sign.
it can be left like that for hours to sign everything automatically

image

PR #513, opened 2021-09-26
Mint NFT from CLI

BertrandBev:
There is no way to mint a regular NFT (out of a candy-machine) from the CLI. Adding that here


Arrowana:
There is this https://github.com/metaplex-foundation/metaplex/blob/master/rust/token-metadata/test/src/main.rs
I would tend to trust a lot more some sound rust code than some typescript CLI. It is in "test" for strange reasons.
Maybe let's move it to `token-metadata/cli` and document it?

BertrandBev:
> I would tend to trust a lot more some sound rust code than some typescript CLI

At the same time, the procedure is rather straightforward (basically just calling rust programs around), and is exactly what both the `candy-machine-cli` & the web client are doing! 
Additionally, there doesn't seem to be a documented rust CLI for now, and it would just seem odd to have two CLIs doing kind of the same thing in different places.
I guess I don't know the reasons why the whole existing CLI hasn't been made in rust in the first place, but now that we have a solid TS solution, it sounds meaningful to complement it with that rather useful command! Don't you think?

Arrowana:
Unfortunately, like you mentioned, it is mostly in typescript, so we are doomed, please add some more.

However, why are things like `MasterEditionV1`, `Edition` copied once more? This is getting to extreme levels of spaghettiness overall.

Same for the instructions, i feel all this should be in the common package. We need to keep the copy/paste under check because this is going to bubble out of control. However, for the instructions it seems like someone started that.

BertrandBev:
> However, why are things like `MasterEditionV1`, `Edition`... copied once more? This is getting to extreme levels of spaghettiness overall.

Yep I wondered why the cli package had no dependency whatsoever on the common package. I assumes it was because, at least at some point, there was a bunch of tsx files containing a bunch of React code that didn't play well with the cli

Arrowana:
Damn the tsx...
Ok, someone one day will clean that up and make it a standalone ts package.

BertrandBev:
Hopefully! And when that happens, we'll shave a bunch from the cli !

js/packages/cli/src/helpers/schema.ts
```diff
@@ -0,0 +1,378 @@
...
+enum MetadataKey {
+  Uninitialized = 0,
+  MetadataV1 = 4,
+  EditionV1 = 1,
+  MasterEditionV1 = 2,
+  MasterEditionV2 = 6,
+  EditionMarker = 7,
+}
+
+class MasterEditionV1 {
```
bhgames, line 18: This should be MasterEditionV2, not V1

bhgames: You're definitely using the wrong struct (the older version) given the master edition creation command being used above. Wont match.

bhgames (changes requested):

js/packages/cli/src/helpers/schema.ts
BertrandBev: Ok, those weren't used anyway, I was maintaining the SHEMA structure used by the front-end
I removed them for now!

bhgames:
Nice. Merged. Send me your wallet.

bhgames:
Also which medals you already have

BertrandBev:
> Nice. Merged. Send me your wallet.

Tanks! 5QLZnmMpT4hZW5hJ5pvHBStdGdgCvGjXidxhAMPQe15m
I got a HoMR 2nd place AGx1 & HoMR 3rd place Proof of History

PR #524, opened 2021-09-27
Improve auctions load by pulling related metadata

m-sebastiyan:
The auction manager has **auction** and **auctionDataExtended** fields in his info. Since the auction manager is being filtered by the store id, we can use these ids to load related auctions.
By having these ids, we can utilize **getMultipleAccounts** to fetch related auctions.
This will clean up the cache and improve loading & rendering speed.

### DIFF
--- a/js/packages/common/src/contexts/meta/loadAccounts.ts
+++ b/js/packages/common/src/contexts/meta/loadAccounts.ts
@@ -305,6 +305,8 @@ export const loadAccounts = async (connection: Connection) => {
     auctionManagersByAuction: state.auctionManagersByAuction,
   }).then(forEachAccount(processAuctions));
 
+  console.log('Metadata size', state.metadata.length);
+
   return state;
 };
 
--- a/js/packages/common/src/contexts/meta/loadAccounts.ts
+++ b/js/packages/common/src/contexts/meta/loadAccounts.ts
@@ -38,6 +38,7 @@ import { getEmptyMetaState } from './getEmptyMetaState';
 import { getMultipleAccounts } from '../accounts/getMultipleAccounts';
 import { getProgramAccounts } from './web3';
 import { createPipelineExecutor } from './createPipelineExecutor';
+import { pullAuctions } from './pullAuctions';
 
 export const USE_SPEED_RUN = false;
 const WHITELISTED_METADATA = ['98vYFjBYS9TguUMWQRPjy2SZuxKuUMcqR4vnQiLjZbte'];
@@ -275,10 +276,6 @@ export const loadAccounts = async (connection: Connection) => {
     getProgramAccounts(connection, VAULT_ID).then(
       forEachAccount(processVaultData),
     );
-  const loadAuctions = () =>
-    getProgramAccounts(connection, AUCTION_ID).then(
-      forEachAccount(processAuctions),
-    );
   const loadMetaplex = () =>
     getProgramAccounts(connection, METAPLEX_ID).then(
       forEachAccount(processMetaplexAccounts),
@@ -298,13 +295,15 @@ export const loadAccounts = async (connection: Connection) => {
   const loading = [
     loadCreators().then(loadMetadata).then(loadEditions),
     loadVaults(),
-    loadAuctions(),
     loadMetaplex(),
   ];
 
   await Promise.all(loading);
 
-  console.log('Metadata size', state.metadata.length);
+  await pullAuctions({
+    connection,
+    auctionManagersByAuction: state.auctionManagersByAuction,
+  }).then(forEachAccount(processAuctions));
 
   return state;
 };
--- a/js/packages/common/src/contexts/meta/pullAuctions.ts
+++ b/js/packages/common/src/contexts/meta/pullAuctions.ts
@@ -1,0 +1,78 @@
+import { Connection } from '@solana/web3.js';
+
+import { getMultipleAccounts } from '../accounts/getMultipleAccounts';
+import { AuctionManagerV1, AuctionManagerV2 } from '../../models';
+import { ParsedAccount } from '../accounts/types';
+import { AccountAndPubkey } from './types';
+
+type AuctionManagerByAuction = Record<
+  string,
+  ParsedAccount<AuctionManagerV1 | AuctionManagerV2>
+>;
+
+interface IPullAuctionsParams {
+  connection: Connection;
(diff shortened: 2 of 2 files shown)


kespinola:
Just completed QA of this change. No bids are displaying for auctions.

kespinola:
Auctions are all showing 1 as the bid.

kespinola:
image

js/packages/common/src/contexts/meta/loadAccounts.ts
```diff
@@ -276,9 +277,10 @@ export const loadAccounts = async (connection: Connection) => {
       forEachAccount(processVaultData),
     );
   const loadAuctions = () =>
-    getProgramAccounts(connection, AUCTION_ID).then(
-      forEachAccount(processAuctions),
-    );
+    pullAuctions({
+      connection,
+      auctionManagersByAuction: state.auctionManagersByAuction,
+    }).then(forEachAccount(processAuctions));
```
kespinola, line 283:
```
      // bidder metadata
      getProgramAccounts(connection, AUCTION_ID, {
        filters: [
          {
            memcmp: {
              offset: 32,
              bytes: auctionManager.info.auction,
            },
          },
        ],
      }).then(forEach(processAuctions)),
      // bidder pot
      getProgramAccounts(connection, AUCTION_ID, {
        filters: [
          {
            memcmp: {
              offset: 64,
              bytes: auctionManager.info.auction,
            },
          },
        ],
      }).then(forEach(processAuctions)),
```

I believe this is what is missing from the auction state now. Current update just getting auctions from the auction program but not bidder pots and bidder metata. I'm looking into adding those queries.

kespinola:
@m-sebastiyan this fixes up the missing bid information. I did lower the chunk size to 20 since getting the bid information is going to be 2 more additional queries for each auction.
https://github.com/metaplex-foundation/metaplex/pull/526/commits/1ed170bb3dd85c178dd6926a87de86e3d1be9af8

m-sebastiyan:
@kespinola bidderMetadata & bidderPots are similar requests. Are you sure they should be duplicated?

```
memcmp: {
              offset: 32,
              bytes: auction.pubkey,
            },
```

```
            memcmp: {
              offset: 32,
              bytes: auction.pubkey,
            },
```

m-sebastiyan:
@kespinola I will align with similar requests from the new JS SDK

I think you have typo in offset. Also, can be optimized by providing data size. 

```
async getBidderPots(connection: Connection) {
    return (
      await AuctionProgram.getProgramAccounts(connection, {
        filters: [
          // Filter for BidderPot by data size
          {
            dataSize: BidderPot.DATA_SIZE,
          },
          // Filter for assigned to this auction
          {
            memcmp: {
              offset: 32 + 32,
              bytes: this.pubkey.toBase58(),
            },
          },
        ],
      })
    ).map((account) => BidderPot.from(account));
  }

  async getBidderMetadata(connection: Connection) {
    return (
      await AuctionProgram.getProgramAccounts(connection, {
        filters: [
          // Filter for BidderMetadata by data size
          {
            dataSize: BidderMetadata.DATA_SIZE,
          },
          // Filter for assigned to this auction
          {
            memcmp: {
              offset: 32,
              bytes: this.pubkey.toBase58(),
            },
          },
        ],
      })
    ).map((account) => BidderMetadata.from(account));
  }
```

js/packages/common/src/contexts/meta/pullAuctions.ts
```diff
@@ -0,0 +1,97 @@
...
+    ),
+  ];
+
+  const auctionsChunks = await Promise.all(
+    auctionIdsChunks.map(ids =>
+      pullAuctionsByAuctionManager({ ids, connection }),
+    ),
+  );
+
+  const bidderMetadata = await getProgramAccounts(connection, AUCTION_ID, {
```
m-sebastiyan, line 39: @kespinola it was getting overloaded by the number of requests in case of getting bidderMetadata & bidderPots for each auction. Left these requests without auction filter

kespinola:
> @kespinola bidderMetadata & bidderPots are similar requests. Are you sure they should be duplicated?
> 
> ```
> memcmp: {
>               offset: 32,
>               bytes: auction.pubkey,
>             },
> ```
> 
> ```
>             memcmp: {
>               offset: 32,
>               bytes: auction.pubkey,
>             },
> ```

Both accounts are saved on the auciton program but the offsets are 32 and 64. Its correct in my initial post and in what you have drafted. I like the additional filter by data size. Can see if rpc responds quicker when it can first toss out all accounts that don't match the data size then filter by the ones for the auction manager.

js/packages/common/src/contexts/meta/pullAuctions.ts
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  const bidderPots = getProgramAccounts(connection, AUCTION_ID, {
+    filters: [
+      // Filter for BidderPot by data size
+      {
+        dataSize: BIDDER_POT_LEN,
+      },
+    ],
+  });
```
kespinola, line 53: This is better than querying all bidder pots and metadata by the auction manager in the init page load. However, as the program grows this will eventually cause processing delays on the frontend as it parses and tosses out accounts not needed for the storefront.

js/packages/common/src/contexts/meta/pullAuctions.ts
kespinola: Yep I encountered the same thing. Agree this is the correct approach until auctions are loaded on demand as the user scrolls the list.

kespinola (changes requested):
Testing notes:

- Bids do not appear when directly visiting an auction

image

- The first photo is the time to request the 2 filtered queries for bid metadata and bidder pots. The second is doing the full auction program fetch. It's just as fast or faster to request all the accounts for programs.

image

image

m-sebastiyan:
@kespinola 
1. Will check the issue, had auctions working fine from my side.
2. It's not only about the request speed, but also about the data amount we transfer, store and process.

SerhiiKoziy:
@kespinola checked this PR, on devnet - works fine for auction, Can you check is it you use devnet?

kespinola:
The issue is on the direct link to the auction. Clicking from the auctions list into the show displays the bids but going straight to it. I agree to send just what is needed over the wire and not processing unneeded accounts is what its all about.

PR #526, opened 2021-09-27
Fix loading auctions in patches

kespinola:
### Issue
The refactor to loading auctions based on auction manager the bidder metadata and bidder pots were no longer being queried.

### Resolution
While doing the chunk load of auctions also query bidder metadata and bidder pots.


kespinola:
Change incorporated into the original PR. closing this one:

https://github.com/metaplex-foundation/metaplex/pull/524

PR #528, opened 2021-09-27
Fast metaplex

bhgames:


kespinola:
In working through performance improvements I believe all the necessary associations for querying relevant accounts by either the owner address or auction manager is sufficient. Storefront apps should query auction managers and auctions upfront but only request associated information like bidder metadata and bidder pots when the auction needs to be displayed to the user. For example, load auction associated info for the first 8 auctions displayed to the user when they first visit the page then request another n as they scroll. 

Once this is in place, the next big improvement is to only query all metadata by creators when the user visits the artworks tab. When visiting the auctions page just query the metadata associated to the auction. This will get the initial set of queries down to the storefront and auction managers. Everything else will be queried dependent on the page the user is visiting.

What would be nice is to have a new version of white-listed creators that has a foreign key to the storefront. Its the only account I encountered that had no means querying using filters.

js/packages/common/src/models/metaplex/index.ts
```diff
@@ -90,6 +95,51 @@ export class PayoutTicket {
...
+    this.page = args.page;
+    this.auctionCaches = args.auctionCaches;
+  }
+}
+
+export class AuctionCache {
+  key: MetaplexKey = MetaplexKey.AuctionCacheV1;
+  store: StringPublicKey;
+  timestamp: BN;
+  metadata: StringPublicKey[];
```
kespinola, line 120: This is nice. help a lot in rendering the auctions page without querying all nfts associated to the creator.

bhgames:
@kespinola That's the plan - change the AuctionView layer to accept idea of being complete without pots and then do a query on load at that next level in show. It's gonna be some gymnastics.

PR #529, opened 2021-09-28
Remove duplicate metadata from state

kespinola:
### Issue
Latest updates to querying metadata by creators introduced duplicate metatda when multiple creators attached to NFTs.

### Changes
- Remove duplicate NFTs before saving by checking pubkey on the account.

https://github.com/metaplex-foundation/metaplex/pull/497


js/packages/common/src/contexts/meta/loadAccounts.ts
```diff
@@ -19,6 +19,7 @@ import {
   decodeMetadata,
   getAuctionExtended,
 } from '../../actions';
+import { uniqWith } from 'lodash';
```
adamjeffries, line 22: I'm not crazy about adding more lodash usage - but I guess there are already debounce and other references

PR #569, opened 2021-10-01
feat: set fixed cost per arweave file

bartosz-lipinski:


aheckmann:
@bartosz-lipinski  Why was this change made? What impact on price is expected? We're seeing reports of large price increases

PR #584, opened 2021-10-02
Validate NFT Cover Images are Over 25KB

kespinola:
### Issue
- we have received reports of files under 25kb resulting in successful uploads to arweave but no content sized to the file. Arweave has notified us that this is an issue with the arweave to sol upload cloud function operated by Metaplex. 

### Resolution
Validate that all cover images for NFTs are larger than 25kb.

Screenshot from 2021-10-02 13-09-44


bhgames:
Love it. @kespinola drop me your wallet

bhgames:
Also which medals you already have

PR #607, opened 2021-10-04
Support for Apple M1 for cli tool

monomadic:
Tested this on an arm macbook air m1 2020 model.


bhgames:
Send me wallet for medal

PR #614, opened 2021-10-04
Fix candy machine

bhgames:


AlexLeung:
Would appreciate some clarification on what happened.

PR #624, opened 2021-10-05
[Fix] Instant sale purchasing

kespinola:
### Background
The instant sale action is 2 transactions. The first for placing a bid and another immediately after for redeeming it.

### Issue
Currently, in between the two transactions, the action calls the `update` function on the meta context which triggers a complete download for all site data. Users would leave the browser session after placing the bid but not redeem it. This left several users with sol placed in escrow but the NFT was not sent to the user.

  ### Resolution
  - Replace the `update` meta context function with one that patches the state given a segment of meta state. A load multiple accounts helper was introduced to query just the accounts that changed from the placing of the bid.
 - The displaying of the redeem button is adjusted to display the button in the case of a bid being placed but not redeemed.

### DIFF
--- a/js/packages/common/src/contexts/meta/loadAccounts.ts
+++ b/js/packages/common/src/contexts/meta/loadAccounts.ts
@@ -1030,7 +1030,7 @@ export const initMetadata = async (
   }
 };
 
-export const loadMultipleAccounts = async (
+export const loadMultipleAccountsIntoMetaState = async (
   conn: Connection,
   keys: StringPublicKey[],
   commitment: string,
--- a/js/packages/common/src/contexts/meta/loadAccounts.ts
+++ b/js/packages/common/src/contexts/meta/loadAccounts.ts
@@ -1029,3 +1029,57 @@ export const initMetadata = async (
     }
   }
 };
+
+export const loadMultipleAccounts = async (
+  conn: Connection,
+  keys: StringPublicKey[],
+  commitment: string,
+): Promise<MetaState> => {
+  const tempCache: MetaState = getEmptyMetaState();
+  const updateTemp = makeSetter(tempCache);
+  const { array } = await getMultipleAccounts(conn, keys, commitment);
+
+  await Promise.all(
+    array.map(async (account, i) => {
+      const pubkey = keys[i];
+
+      // account has an incorrect type ascription
+      if (!account) {
+        console.warn(`Didn't see account for pubkey ${pubkey}`);
+
+        return;
+      }
+
+      const PROGRAM_IDS = programIds();
+      const pair = { pubkey, account };
+
+      // account.owner ALSO has an incorrect type ascription
+      const owner =
+        account.owner instanceof PublicKey
+          ? account.owner.toBase58()
+          : (account.owner as string);
+
+      switch (owner) {
+        case PROGRAM_IDS.metadata:
+          await processMetaData(pair, updateTemp);
+          break;
+        case PROGRAM_IDS.vault:
+          await processVaultData(pair, updateTemp);
+          break;
+        case PROGRAM_IDS.auction:
+          await processAuctions(pair, updateTemp);
+          break;
+        case PROGRAM_IDS.metaplex:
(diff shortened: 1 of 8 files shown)


js/packages/common/src/contexts/meta/loadAccounts.ts
```diff
@@ -514,3 +515,57 @@ export const initMetadata = async (
     }
   }
 };
+
+export const loadMultipleAccounts = async (
```
kespinola, line 519: This is a helper function that takes a list of account public keys, calls get multiple accounts, processes the account based on its owner, and returns a meta state snippet that can be merged into the current state.

js/packages/common/src/contexts/meta/meta.tsx
```diff
@@ -81,6 +84,14 @@ export function MetaProvider({ children = null as any }) {
     }
   }
 
+  const patchState: MetaContextState['patchState'] = temp => {
+    const newState = merge({}, state, temp);
+    newState.store = temp.store ?? state.store;
```
kespinola, line 89: The store is the only non-collection value tracked by meta state so needs to default to the previous state if not in the patch state snippet.

js/packages/web/src/components/AuctionCard/index.tsx
```diff
@@ -254,13 +255,14 @@ export const AuctionCard = ({
...
+
   if (
-    (auctionView.isInstantSale &&
-      Number(auctionView.myBidderPot?.info.emptied) !== 0 &&
-      isAuctionManagerAuthorityNotWalletOwner &&
-      auctionView.auction.info.bidState.max.toNumber() === bids.length) ||
-    auctionView.vault.info.state === VaultState.Deactivated
+    !outstandingBid &&
+    auctionView.isInstantSale &&
+    auctionView.myBidderPot?.info.emptied
```
kespinola, line 265: This is a spot worth explicit review. Getting the button to show was a bit of a struggle.

js/packages/common/src/contexts/meta/loadAccounts.ts
bhgames: My only issue here is that there already is a helper called getMultipleAccounts in common. So may want to possibly rename this.

js/packages/web/src/components/AuctionCard/index.tsx
bhgames: I think the issue here is this isAuctionManagerAuthorityNotWalletOwner is that when in cases where the auction manager is viewing the auction, you want to show the button for reclaiming the nfts. I think for instant sale auctions it's a bit strange, because for an auction manager even you kind of what to not show the buttons if they are acting as a customer until after the auction ends and THEN show the reclaim items. So maybe there is a little if inside here where IF you are the auction manager, and IF the auction is over, then show the reclaim items, as oyu would expect, even fi you the auction manager have already bought something. It's confusing.

bhgames (changes requested):

kespinola:
A note that I will be removing the update function from the meta context. With the change, the patch contains the updates that need to be applied to the redeem bid. The update will be swapped with patchState which takes a meta state snippet and merges it into the existing meta state. 

The reason for the flaky instant sale action is update loads the entire site which leaves lots of opportunity for the place bid to set but not redeem. Mainly people leaving their session or changes pages before its complete.

kespinola:
@m-sebastiyan @bhgames this is ready for another review. I removed our adjustments to hiding actions since other changes were added to upstream that correctly showed redeem button.

I tested by creating an instant sale auction and a normal "sell an existing item" auction. For instant sales by 2 collector wallets were able to place and claim the NFT. Other auction type continued to work as before accepting bids and then letting winner redeem the bid.

A few screenshots of me snatching some instant sale NFTs on devnet.

Screenshot from 2021-10-14 11-37-51

Screenshot from 2021-10-14 11-37-35
Screenshot from 2021-10-14 11-37-24
Screenshot from 2021-10-14 11-35-41

PR #627, opened 2021-10-05
fix winner index value

iamoperand:
Looks like there is a bug in computing `winnerIndex` right now. 
And because of that there is an issue in computing `bidRedemptions` for the `AuctionManager`.

Since I wasn't sure about the fix, I went ahead and found the [corresponding logic](https://github.com/metaplex-foundation/metaplex/blob/master/rust/metaplex/program/src/state.rs#L1487) in the rust codebase (and substituted with that).


bhgames:
Nice f-ing catch. Got you down for another medal.

PR #640, opened 2021-10-06
Update nft metadata

BertrandBev:
Handy function to update NFT's metadata

### DIFF
--- a/js/packages/cli/src/helpers/instructions.ts
+++ b/js/packages/cli/src/helpers/instructions.ts
@@ -64,6 +64,7 @@ export function createAssociatedTokenAccountInstruction(
   });
 }
 
+
 export function createMetadataInstruction(
   metadataAccount: PublicKey,
   mint: PublicKey,
@@ -125,6 +126,7 @@ export function createMasterEditionInstruction(
   updateAuthority: PublicKey,
   txnData: Buffer,
 ) {
+  
   const keys = [
     {
       pubkey: editionAccount,


YogaaDev:
Hi, with these modifications we can modify the nfts (images) after being mint? thank you

studentofcoding:
> > Hi, with these modifications we can modify the nfts (images) after being mint? thank you
> 
> Yes, exactly!

Awesome work mate!

Will it be bulk update or one by one?

BertrandBev:
> Will it be bulk update or one by one?

One by one as-is, but it's a really cheap & fast program call, so looping on this function is a no brainer!

YogaaDev:
> > Will it be bulk update or one by one?
> 
> One by one as-is, but it's a really cheap & fast program call, so looping on this function is a no brainer!

Thanks for your response, may I have a presicion.

If I understand your program correctly, it can update the metadata of a key in base58. What does this key correspond to?

   .option ('- m, --mint <string>', 'base58 mint key')
    .option ('- u, --url <string>', 'metadata url')

Also what does this URL metadata mean?

I am a developer but still very new to learning Solana, I need more details ...

Basically if I understand correctly, I have to upload again "metadata" on the blockchain, then come and link these metadata to the mint key. How to create these new metadata?

According to the tutorials that I saw and followed, we create this data with the images and then launched the "upload" command which is responsible for putting them on arweave ...

BertrandBev:
> Drop wallet for medal

@bhgames 5QLZnmMpT4hZW5hJ5pvHBStdGdgCvGjXidxhAMPQe15m
Thanks!

PR #668, opened 2021-10-08
Allow seed authority to seed metadata for existing tokens

hendrikhofstadt:
This will be an enabler for wider adoption of spl-metadata for fungible tokens. Given that Phantom now supports rendering spl-metadata and many of the existing tokens don't have a mint authority it will be beneficial to have the contract authority retroactively populate token meta.

This specifically currently does not allow updating metadata to prevent misuse.


adamjeffries:
@hendrikhofstadt please reach out to me.  I'd like to verbally discuss anything related to modifying token metadata

PR #692, opened 2021-10-11
Adds a tool to deploy all contracts to a local validator

austbot:
This PR adds a tool to deploy all contracts to a local validator for quick and easy local rust development.
It is very helpful to have this on a loop while changing things about the contracts.

### DIFF
--- a/rust/package.json
+++ b/rust/package.json
@@ -17,6 +17,7 @@
   },
   "scripts": {
     "idl": "node test/idlToTs",
+    "deploy-all": "node tools/deployAll.js",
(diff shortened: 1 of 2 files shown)


thlorenz:
This is solved via local validator initialization that adds those programs at validator startup.
This is a lot faster than deploying after startup and also allows using the same pubkeys used on dev/mainnet.

It will get merged soon, for more info see [here](https://github.com/thlorenz/metaplex/blob/73d098870412ecc7566de240e7fc7647ff5bcbfd/tests/setup/src/init-test-validator.ts#L53-L60) for now.

PR #706, opened 2021-10-13
Feature/rust cli

CalebEverett:
I thought it would be good to have to have a full featured Rust cli for the metaplex programs with the same robustness as the spl-token cli. Before getting too far along, I wanted to see if you would like to include a Rust cli as a feature and if so get any feedback you might have on the set up.


aheckmann:
Good stuff @CalebEverett. Before this lands in the foundation, we'd like you to publish the project under your own github username/org and then add a link on our [community docs](https://docs.metaplex.com/community) page. If/when the project gains traction, we can revisit graduating it here.

CalebEverett:
Awesome, I actually have a better project that would be more useful
immediately that  gets assets uploaded to Arweave super efficiently using
their new bundles format.

https://github.com/CalebEverett/arloader

CalebEverett:
I had actually intended to build the metaplex CLI but got down the rabbit
hole on arweave because it was so hard to get assets uploaded there. If you
wanted to promote the metaplex CLI which would be great I'd love an
opportunity to incorporate arloader and make the initial version of
metaplex CLI a bit more turnkey before you put it out for people to start
noticing.

aheckmann:
up to you. just let us know!

CalebEverett:
I'll get an update done and ping you next week

PR #709, opened 2021-10-13
Add EndAuction button for unlimited auctions

n00m4d:


kespinola:
Helping out with some QA. Assuming the auction program changes are deployed to devnet but ending the sale fails but still shows congratulations. 

As a user I expect the error to be captured and displayed to the user.


I did correctly get the end auction option on an active instant sale when logged in as the storefront owner.

image

m-sebastiyan:
@kespinola Please use these program ids

export const AUCTION_ID =
  '4zvwmWY1P5foesnEJo8FsG91RFdsAk5wgKjuMEkKyjVJ' as StringPublicKey;

export const METAPLEX_ID =
  '8sSkg1dMoB6dBPeyHbLcunH2B6Dt6n1MGsUa6nbCms75' as StringPublicKey;

PR #716, opened 2021-10-13
Refactor: use `anchor.BN` instead of `bn.js` in Metaplex CLI

junhoyeo:
- `@project-serum/anchor` has it's version of `bn.js` itself, so there's no need to add the dependency again
  - In case the `BN` implementation of Anchor SDK change, it's better to couple it to the module
  - Suggestion for good first issue: we can clean deps from other packages too

https://github.com/project-serum/anchor/blob/a845af21af936a4278c955f06ee783715e443fc4/ts/src/index.ts#L1

### DIFF
--- a/js/packages/cli/package.json
+++ b/js/packages/cli/package.json
@@ -47,7 +47,6 @@
     "@project-serum/anchor": "^0.14.0",
     "@solana/spl-token": "^0.1.8",
     "arweave": "^1.10.16",
-    "bn.js": "^5.2.0",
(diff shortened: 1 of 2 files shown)


junhoyeo:
> pls drop wallet for medal

@bhgames Wow cool! Solana wallet is `HJLQd7CxQK5epNLE3R4u8b2MdGzmcvXjpntGWfht4FZH`
btw thanks to the maintainers for this awesome protocol

PR #720, opened 2021-10-13
Auction Status for Instant Sales

kespinola:
### Goal

Auction status on the listing (ie auction) view supports instant sale without breaking status on standard auctions.

### Changes
- Migrate status computing to hook reserved for auction rendered card. The hook returns status and amount.

Screenshot from 2021-10-13 15-22-09


m-sebastiyan:
Checked it locally, didn't find any issues.

PR #721, opened 2021-10-14
Allow assets starting from any number

oboxodo:
This change removes the need for asset and metadata files to start with 0 (zero). So instead of having:

```
0.png
0.json
1.png
1.json
2.png
2.json
```

...I can start with 1 (or any number, FWIW) like this:

```
1.png
1.json
2.png
2.json
3.png
3.json
```

This also allows me to generate, say, 100 files names 1.png to 100.png, then split them in different directories, then upload them with one cache file per directory, and then generate one candy machine per directory and cache file. All using different groups of assets from the same generated collection. Before this change, I would've needed to rename all the files and edit all the json metadata after generating them so that each group starts with 0.

### DIFF
--- a/js/packages/cli/src/commands/upload.ts
+++ b/js/packages/cli/src/commands/upload.ts
@@ -178,11 +178,11 @@ export async function upload(
               const index = keys[i];
               return cacheContent.items[index]?.onChain || false;
             });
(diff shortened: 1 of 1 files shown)


aheckmann:
closing, no response. if this is still necessary please reopen and include a video as requested

PR #725, opened 2021-10-14
allow metadata updates where the update authority is not a creator and creators are not being changed

levicook:
The goal of this PR is to enable an update authority to:
- preserve existing creators, while:
- updating the metadata name, symbol, and uri without setting creators to None

This came up on a project where:
- some name properties were duplicated inside of a collection.
- the update authority is not in the creators array

There's probably a case to be made for ensuring creators + seller_fee_basis_points can't be changed. Feedback on this would be greatly appreciated.


bhgames:
Oh yeah this has been fixed now, no longer needed

PR #733, opened 2021-10-15
Drop Using An Existing URL on Creating NFT

kespinola:
### Changes
- Adding an image for NFT using an existing URL (instead of file upload) does not work with uploadFile4 so its been removed.
- Fix issue with alert persisting in the state causing the user to always see a failure message even after correcting the error.
- Adjust steps progress to be more aligned with the execution of tasks required to create the NFT.
- Print contents of error messages from sol to arweave upload.

image


aheckmann:
Is this one still needed? Will close this one due to merge conflicts. Please reopen with clean up if necessary.

PR #753, opened 2021-10-18
fix fun bug

staccDOTsol:
This fixes Jordans fun bug.

Update authorities not in the creators array will now again be able to make changes to the creators array along with the rest of the data struct, when passed.

I may be naive about the original intent surrounding the 'found' variable and constraints,  feel free not to merge this.


staccDOTsol:
@bhgames conversationally should we just let an updateauthority update the creators array anyways?

I know you outlined the 'real' fix which I'm willing to submit, however

considering the people unintentionally exploiting may well need this, and assumed they could do it, cuz they were

this PR is probably the superior solution.

just let the updateauthorities do what they want

aheckmann:
Thank you, this one was fixed a couple weeks ago. Closing

Issue #659, opened 2021-10-07
Improving the Next.js Developer Experience


dankmemetics:
Hello friends from Metaplex,

Recently, I've forked Metaplex and created this project: https://dankexotics.cloud/
However, there were several comments and constructive feedback I had about the Next.js codebase.

----

#### Use Next.js routing

`react-router-dom` was integrated into the source code. This is not necessary when the entire source can be based on `Next.js` routing.

See [dynamic routes](https://nextjs.org/docs/routing/dynamic-routes) in Next.js to learn more about how that works.

#### Asynchronous React Hooks

This is actually a HUGE problem and anti pattern in the source code. Using asynchronous hooks **outside of `useEffect` or `useLayoutEffect`** will cause React to rerender the DOM several times over needlessly. This actually causes the webpage to lock up. The webpage actually often becomes unresponsive because of this.

A better solution would be to update state in Redux. Given the nature of the project. It would be more ideal and easier to share state via Redux as opposed to having state contained in single state components or parent components.

#### Using `@oyster/common`

I understand why `@oyster/common` was used as a way to share dependencies between folders... But I don't see how doing an import such as:

```typescript
import { } from '../../common/component';
```

Wouldn't have achieved the same effect. Since with the `tsconfig.json` file you can just add folders to the `include` dependency array.

----

Hopefully this feedback was helpful.

I would be more than willing to help contribute a complete refactor of the Next.js codebase when we have a v1 RC of the [Metaplex JS Library](https://github.com/metaplex/js).

If anything, when that library is released. Happy to create a new repo independently and send in a refactor for review.

dankmemetics:
@jordansexton I think it'd be best I invest time into that once the JS library is fully operational.

It'll take longer til execution time. But will 100% reduce amount of scope creep in any related PRs.

PR #762, opened 2021-10-19
Use Nextjs router

josezy:
I started moving to nextjs router but [dynamic routing](https://nextjs.org/docs/routing/dynamic-routes) is not working, simply not matching the URL, I haven't found the problem so I need help @dankmemetics @jordansexton 

fixes part of #659


dankmemetics:
Hey! @josezy, I just saw this..
Sorry for the latency (2 1/2 weeks).

I will review it this weekend.

PR #766, opened 2021-10-20
feat: calculate proper cost to store

zaxozhu:
Goal: Calculate the proper cost to store the asset, instead of using hardcoded lamports value.


aheckmann:
closing in favor of #773

PR #767, opened 2021-10-20
Fixed #724: Metaplex logo not visible in the readme

### LINKED ISSUE #724: Metaplex logo not visible in the readme
Readme.md

ptisserand:
Minor modification to update img url

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
   <a href="https://metaplex.com">
-    <img alt="Metaplex" src="https://metaplex.com/meta.svg" width="250" />
+    <img alt="Metaplex" src="https://metaplex.com/static/logos/metaplex.svg" width="250" />
   </a>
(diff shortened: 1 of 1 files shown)


thlorenz:
Hi, thanks for the fix, but it seems like that the logo isn't at the URL you provided anymore either.

I found it at https://assets.website-files.com/6182ee30b608385a15466a3f/6184397712e6442ff28e7533_metaplex_logo.svg, but judging from the URL this might change in the future. BTW to see it you need to set its background to black like in the below screenshot.

Screen Shot 2021-11-22 at 8 17 20 PM

I think a better solution might be to download the SVG file and add it to an `./assets` folder at the root of the project and then link directly from there, similarly to how I did in [this example](https://github.com/thlorenz/phan-wallet-mock-react-example#phan-wallet-mock-app-)

PR #777, opened 2021-10-22
Added Whitelist from File option

nftinkerer:
Guarantees FLP lottery win to those in an FLP json flie.

### DIFF
--- a/js/packages/cli/src/fair-launch-cli.ts
+++ b/js/packages/cli/src/fair-launch-cli.ts
@@ -2326,7 +2326,7 @@ program
 
     const ticketsFlattened = ticketKeys.flat();
 
-    const states: { seq: number; number: anchor.BN; eligible: boolean, whitelist: boolean }[][] =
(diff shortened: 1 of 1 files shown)


AnderUstarroz:
This took me off-ward, so the white list kind of gives priority to white listed users but actually won't prevent users that are not in the list to buy?
Example:

```
More or equal nfts than winners, everybody wins.
Lottery results [
  {
    seq: 0,
    number: 500000000,
    eligible: true,
    whitelisted: true,
    chosen: true
  },
  {
    seq: 1,
    number: 450000000,
    eligible: true,
    whitelisted: false,
    chosen: true
  }
]
```

AnderUstarroz:
Just a suggestion,  converting the white list array into an object makes the search of items dramatically faster. That would be using:

```
 el.buyer.toBase58() in whitelist
 ```
 Instead of:
 ```
whitelist.includes(el.buyer.toBase58())
```

[Try running this](https://jsbench.me/irkhdxnoqa/2) to see the huge difference in performance specially when having thousands of rows (like is normally the case with NFTs).

PR #787, opened 2021-10-24
Auction caches pushed at the corresponding offset

Arrowana:
Problem:
- When populating the store indexer the first time some auctions did not make it while they were AuctionManagerV2
- When creating new auctions, sometimes an auction cache is created but somehow the push to the store indexer failed (something else to be fixed)
https://explorer.solana.com/tx/5di1ucsrkrMAxyH569kDrP1VGZy4PpZHGja4fH9rFHXS2x1kw186rQAWPPrqyFjXftsf2N9vEPWbZv2FrKVTqTSL
I suspect the auction cache is out of whack after the first push? Maybe not being updated

This does not fix the leak but this provides a way to save anything that has fallen through.

Solution: If a transaction fails and auction disappears, the magic button can save those

### DIFF
--- a/js/packages/web/src/actions/cacheAllAuctions.ts
+++ b/js/packages/web/src/actions/cacheAllAuctions.ts
@@ -1,6 +1,5 @@
-import { Connection, Keypair, TransactionInstruction } from '@solana/web3.js';
+import { Connection } from '@solana/web3.js';
 import {
-  getAuctionCache,
   loadAccounts,
   MetaplexKey,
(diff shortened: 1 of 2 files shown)


bhgames:
LMK when ready, just reviewed it, looks fine (Except for fact that it appears to only work for auctions on page 0)

bhgames (approved):

bhgames:
this looks good, if you can finish it and undraft it i'll merge it

PR #803, opened 2021-10-26
Add creator parameter to create_metadata_account

jnuno98:
The create_metadata_account function added None creator as a default value. I have added a parameter to the function call that lets the user add a new creator to the token metadata.

### DIFF
--- a/rust/token-metadata/test/src/main.rs
+++ b/rust/token-metadata/test/src/main.rs
@@ -9,7 +9,7 @@ use {
             update_metadata_accounts,
         },
         state::{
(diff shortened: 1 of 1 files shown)


rust/token-metadata/test/src/main.rs
```diff
@@ -583,7 +594,7 @@ fn create_metadata_account_call(
         name,
         symbol,
         uri,
-        None,
+        Some(creators),
```
austbot, line 597: please handle the case if address/creator not passed in

austbot (changes requested):

PR #804, opened 2021-10-26
use correct share value for creators

yamijuan:


austbot (approved):
This is a great contribution

austbot (approved):

js/packages/web/src/hooks/useCreators.ts
```diff
@@ -20,18 +20,18 @@ export const useCreators = (auction?: AuctionView) => {
         )
           .filter(creator => creator.verified)
           .reduce((agg, item) => {
-            agg.add(item.address);
+            agg.add([item.address, item.share]);
             return agg;
-          }, new Set<string>())
+          }, new Set<[string, number]>())
```
aheckmann, line 25: The purpose of the reduce is to dedup the creators. This change will prevent the Set from deduplicating bc each array literal is a separate object.

To retain the deduplication functionality, change the Set to a Map<string, number> using the address as the key and share as the value

aheckmann (changes requested):

austbot (approved):

js/packages/web/src/hooks/useCreators.ts
```diff
@@ -20,18 +20,18 @@ export const useCreators = (auction?: AuctionView) => {
         )
           .filter(creator => creator.verified)
           .reduce((agg, item) => {
-            agg.add(item.address);
+            agg[item.address] = [item.address, item.share];
```
austbot, line 23: @aheckmann good?

aheckmann: {}. values () is not a function
change it to a Map

yamijuan: I'll do it

aheckmann (approved):

PR #810, opened 2021-10-27
fair launch candy ui

staccDOTsol:
it's a fair launch UI

the only change is that if it has a candy machine Id but no fair launch Id, it skips to Stage4


staccDOTsol:
@bhgames this does not work afaik. I may have just been minted out of my 4 candy machine by bots tho. I will revisit tomorrow

bhgames:
really nice. I'll merge this if you can sho me a video of an E2E quick fair launch still working AND a candy machine by itself working.

PR #818, opened 2021-10-28
Fix Link

jdcaballerov:


ARCHITECTURE.md
```diff
@@ -178,7 +178,8 @@ pub struct EditionMarker {
 
 ```
 
-The instruction set for the token metadata contract can be found here: [https://github.com/metaplex-foundation/metaplex/blob/master/rust/token-metadata/program/src/instruction.rs](https://github.com/metaplex-foundation/metaplex/blob/master/rust/token-vault/program/src/instruction.rs)
+The instruction set for the token metadata contract can be found here: [https://github.com/metaplex-foundation/metaplex/blob/master/rust/token-metadata/program/src/instruction.rs](https://github.com/metaplex-foundation/metaplex/blob/master/rust/token-metadata/program/src/instruction.rs)
+
```
jordansexton, line 182:
```suggestion
The instruction set for the token metadata contract can be found here: https://github.com/metaplex-foundation/metaplex/blob/master/rust/token-metadata/program/src/instruction.rs
```

Why alias the link when we have the whole URL?

PR #828, opened 2021-10-29
Native Arweave upload

21e8:
In case users do have the option to purchase Arweave it is now possible to pass a wallet file along and upload directly.


21e8:
@bhgames see https://github.com/metaplex-foundation/metaplex/pull/828
Shall we close in favor of that branch? then I will provide video

bhgames:
Sure, I'll check the other one out

PR #838, opened 2021-10-29
refactor: use hook on auctions list

m-sebastiyan:
### In this PR

Moved auction view's data processing to a hook.

#### How it worked before
**Arrays generation**
Auction view arrays are generated every time a component re-renders without any limitations. Even though it needs to show live auctions, it still generates **participated, resaleAuctions, auctionsEnded**

Old code:
```ts
  // Arrays for every type are generated above and then chosed by this switch-case.
  switch (activeKey) {
    case LiveAuctionViewState.All:
      items = liveAuctions;
      break;
    case LiveAuctionViewState.Participated:
      items = participated;
      break;
    case LiveAuctionViewState.Resale:
      items = resaleAuctions;
      break;
    case LiveAuctionViewState.Ended:
      items = auctionsEnded;
      break;
  }
```

**Rendering function**
Each time component renders, two rendering functions are re-creating **liveAuctionsView** and **endedAuctions**. They use related auctions arrays. 

##### Arrays generation
There is only one array generated based on a selected tab. It's also being memoed. 

```ts
  const auctions = useAuctions();

  const filteredAuctions = useMemo(() => {
    const filterFn = getFilterFunction(activeKey);

    return sortAuctionsByDate(auctions).filter(auction =>
      filterFn(auction, publicKey),
    );
  }, [activeKey, auctions, publicKey]);
```

##### Rendering
Content rendering is not spread out between tabs. There is only one place it's being rendered.

```tsx
<Row>
  <Masonry
    breakpointCols={breakpointColumnsObj}
    className="masonry-grid"
    columnClassName="masonry-grid_column"
  >
    {isLoading &&
      [...Array(10)].map((_, idx) => <CardLoader key={idx} />)}
    {!isLoading &&
      auctions.map((m, idx) => (
        <Link to={`/auction/${m.auction.pubkey}`} key={idx}>
          <AuctionRenderCard
            key={m.auction.pubkey}
            auctionView={m}
          />
        </Link>
      ))}
  </Masonry>
</Row>
```

### DIFF
--- a/js/packages/web/src/styles/app.less
+++ b/js/packages/web/src/styles/app.less
@@ -280,3 +280,34 @@ model-viewer {
     padding: 15px;
   }
 }
+
+.my-masonry-grid {
+  display: flex;
+  width: 100%;
+  margin-left: auto;
+  margin-right: auto;
+  margin-left: -32px !important;
+}
+
+.my-masonry-grid_column {
+  padding-left: 32px !important; /* gutter size */
+  background-clip: padding-box;
+}
+
+.my-masonry-grid_column > div {
+  /* change div to reference your elements you put in <Masonry> */
+  margin-bottom: 30px;
+}
+
+/* Optional, different gutter size on mobile */
+@media (max-width: 800px) {
+  .my-masonry-grid {
+    margin-left: -15px; /* gutter size offset */
+  }
+  .my-masonry-grid_column {
+    padding-left: 15px; /* gutter size offset */
+  }
+  .my-masonry-grid_column > div {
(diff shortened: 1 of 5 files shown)


kurpav (approved):

austbot (commented):
This is looking good, let me pull and run

kespinola (approved):
looks good to me. Holaplex has a modification to the auction list that lazy loads bids as you scroll. Shooting to get a PR into upstream end of next week.

austbot (approved):

PR #843, opened 2021-10-30
Fix AUCTION_HOUSE_SIZE calculation

RohanKapurDEV:
`AUCTION_HOUSE_SIZE` seems to allocate 8 bytes for `u8`s and nothing at all for `can_change_sale_price`. The error itself is all covered by the extra padding but we may as well fix it up properly so its not confusing to read

### DIFF
--- a/rust/auction-house/src/lib.rs
+++ b/rust/auction-house/src/lib.rs
@@ -1345,11 +1345,12 @@ pub const AUCTION_HOUSE_SIZE: usize = 8 + //key
 32 + //treasury mint
 32 + //authority
 32 + // creator
-8 + // bump
-8 + // treasury_bump
-8 + // fee_payer_bump
+1 + // bump
+1 + // treasury_bump
+1 + // fee_payer_bump
 2 + // seller fee basis points
 1 + // requires sign off
+1 + // can change sale price
 200; //padding
 
 #[account]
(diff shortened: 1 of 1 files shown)


rust/auction-house/src/lib.rs
```diff
@@ -1345,11 +1345,12 @@ pub const AUCTION_HOUSE_SIZE: usize = 8 + //key
 32 + //treasury mint
 32 + //authority
 32 + // creator
-8 + // bump
-8 + // treasury_bump
-8 + // fee_payer_bump
+1 + // bump
+1 + // treasury_bump
+1 + // fee_payer_bump
```
bhgames, line 1350: good catch, but to keep this all the same size we need to move the extra 21 bytes to the padding.

RohanKapurDEV: Addressed by 6affeef0d24b158f062e389ed00c5ad9c47aa83b

rust/auction-house/src/lib.rs
```diff
@@ -1345,12 +1345,13 @@ pub const AUCTION_HOUSE_SIZE: usize = 8 + //key
...
 2 + // seller fee basis points
 1 + // requires sign off
-200; //padding
+1 + // can change sale price
```
bhgames, line 1353: I think we're off by one. You added 1, and 21, and subtracted 21. 22 - 21 = 1. Needs to be 0.

RohanKapurDEV: so padding should be 220 right?

RohanKapurDEV:
Everything should be fine to merge for now I think

austbot (approved):

PR #881, opened 2021-11-02
chore: fix token-meta tests+clean up code

thlorenz:
This PR focuses mostly on fixing token-metadata tests that were broken.
As part of this build errors needed to be fixed in the auction tests.

NOTE: the formatting changes and _in module wrapping_ create a lot of noise especially in the tests, therefore to focus on actual changes to the tests please refer to this https://github.com/metaplex-foundation/metaplex/pull/881/commits/2e9431f7066f4ace4d59815730b045a1c3f86309

The main fix entailed expecting strings to be _puffed out_ to have a deterministic length.
Some docs were added around and a method factored out to be useable from both the program
itself as well as the tests.

I also wrapped each test file in a module in order to be able to run them in isolation + get improved failure/success output, see https://github.com/metaplex-foundation/metaplex/pull/881/commits/14a147e995af5825ec14df0f6d657476a6b71687.

Since the imports were very unorganized I added a rustfmt at the end and reformatted all code
to match it.

I chose `max_width = 100` since that matched the existing code, however if we want to change
this to `80` instead I can update the PR.
If these style changes shouldn't be part of this at all I can back those commits out and we can
add them in a later PR.

### Follow Up

- next I'll get these tests to run in CI an pass
- after that I'll look through the contract and existing tests in more detail and will add more tests were needed (right now one success test and only to failure case tests exist for each instruction)

### DIFF
--- a/rust/token-metadata/program/src/deprecated_instruction.rs
+++ b/rust/token-metadata/program/src/deprecated_instruction.rs
@@ -1,14 +1,12 @@
-use {
-    crate::{
-        instruction::{CreateMasterEditionArgs, MetadataInstruction},
-        state::Reservation,
-    },
-    borsh::{BorshDeserialize, BorshSerialize},
-    solana_program::{
-        instruction::{AccountMeta, Instruction},
-        pubkey::Pubkey,
-        sysvar,
-    },
+use crate::{
+    instruction::{CreateMasterEditionArgs, MetadataInstruction},
+    state::Reservation,
+};
+use borsh::{BorshDeserialize, BorshSerialize};
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+    sysvar,
 };
 
 #[repr(C)]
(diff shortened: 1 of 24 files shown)


rust/token-metadata/program/src/deprecated_instruction.rs
```diff
@@ -1,14 +1,12 @@
...
-        instruction::{CreateMasterEditionArgs, MetadataInstruction},
-        state::Reservation,
-    },
-    borsh::{BorshDeserialize, BorshSerialize},
-    solana_program::{
-        instruction::{AccountMeta, Instruction},
-        pubkey::Pubkey,
-        sysvar,
-    },
+use crate::{
```
thlorenz, line 1: I'm realizing that here we end up with multiple `use` statements (this is how it was in tests, just less organized).
We should agree on one approach and if it is different than the code is at this point I can adapt the `rustfmt.toml` and this PR to respect it.

rust/token-metadata/program/src/utils.rs
```diff
@@ -898,26 +898,21 @@ pub fn process_create_metadata_accounts_logic(
...
-        array_of_zeroes.push(0u8);
-    }
-    metadata.data.symbol =
-        metadata.data.symbol.clone() + std::str::from_utf8(&array_of_zeroes).unwrap();
+    metadata.data.name = puffed_out_string(&metadata.data.name, MAX_NAME_LENGTH);
+    metadata.data.symbol = puffed_out_string(&metadata.data.symbol, MAX_SYMBOL_LENGTH);
+    metadata.data.uri = puffed_out_string(&metadata.data.uri, MAX_URI_LENGTH);
+}
 
+pub fn puffed_out_string(s: &String, size: usize) -> String {
```
aheckmann, line 909: nice

aheckmann (approved):

PR #892, opened 2021-11-03
Adding parameter for rent reduction

MarkSackerberg:
Even though it will be possible to reclaim rent soon in the metaplex discord currently many people are asking on how to save costs. And one thing that many are doing now is reduce rent cost to something lower than rent exemption amount.

To make it easier for them and do not have the same discussions over and over where to modify the CLI I've added this new parameter.

### DIFF
--- a/js/packages/cli/src/helpers/instructions.ts
+++ b/js/packages/cli/src/helpers/instructions.ts
@@ -210,6 +210,7 @@ export async function createConfigAccount(
   configData,
   payerWallet,
   configAccount,
+  rentTimeInDays,
 ) {
   const size =
     CONFIG_ARRAY_START +
@@ -218,14 +219,25 @@ export async function createConfigAccount(
     4 +
     Math.ceil(configData.maxNumberOfLines.toNumber() / 8);
 
+  let rent = await anchorProgram.provider.connection.getMinimumBalanceForRentExemption(
+    size,
+  );
+  
+  // calculate rent
+  if (typeof rentTimeInDays !== 'undefined') {
+    console.log('Instead of paying ' + rent + ' Lamports for rent')
+    let rentMultiplier = rentTimeInDays / 30;
+    rent = rent / 24 * rentMultiplier;
+    console.log('only paying ' + rent  + ' Lamports for rent')
+  } else {
+    console.log('paying ' + rent + ' Lamports for rent')
+  }
+  
   return anchor.web3.SystemProgram.createAccount({
     fromPubkey: payerWallet,
     newAccountPubkey: configAccount,
     space: size,
-    lamports:
-      await anchorProgram.provider.connection.getMinimumBalanceForRentExemption(
-        size,
-      ),
+    lamports: rent,
     programId: CANDY_MACHINE_PROGRAM_ID,
   });
 }
--- a/js/packages/cli/src/candy-machine-cli.ts
+++ b/js/packages/cli/src/candy-machine-cli.ts
@@ -77,6 +77,10 @@ programCommand('upload')
     '-r, --rpc-url <string>',
     'custom rpc url since this is a heavy command',
   )
+  .option(
+    '-d, --rent-days <number>',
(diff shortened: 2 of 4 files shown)


Qambar:
Great work @MarkSackerberg, I am going to test this out with one of the projects that I am launching.

js/packages/cli/src/helpers/instructions.ts
```diff
@@ -218,14 +219,25 @@ export async function createConfigAccount(
...
 
+  let rent = await anchorProgram.provider.connection.getMinimumBalanceForRentExemption(
+    size,
+  );
+  
+  // calculate rent
+  if (typeof rentTimeInDays !== 'undefined') {
+    console.log('Instead of paying ' + rent + ' Lamports for rent')
+    let rentMultiplier = rentTimeInDays / 30;
+    rent = rent / 24 * rentMultiplier;
```
bhgames, line 230: where are you getting this math from?

MarkSackerberg: Hi Jordan,
you are right, it is not correct. Looks like i commited a WIP. 
What we really would need to do is
// First get the number of epochs. +1 to be save in case the epoch ends in the next minutes.
const epochs = Math.ceil(rentTimeInDays/2) + 1;

// get the rent per byte. There is no implementation in web3 for it right now. To get an approximation divide the cost of rent exemption with the appoximate number of epochs in two years
const rent_per_byte_epoch = rent / 365;

// and finally multiply our count of epochs with the rent.
rent = epochs * rent_per_byte_epoch

KartikSoneji: We can also use the calculation described in the docs:
> As of writing, the fixed rent fee is 19.055441478439427 lamports per byte-epoch on the testnet and mainnet-beta clusters. An epoch is targeted to be 2 days (For devnet, the rent fee is 0.3608183131797095 lamports per byte-epoch with its 54m36s-long epoch).
>
> This value is calculated to target 0.01 SOL per mebibyte-day (exactly matching to 3.56 SOL per mebibyte-year):
>
> ```text
> Rent fee: 19.055441478439427 = 10_000_000 (0.01 SOL) * 365(approx. day in a year) / (1024 * 1024)(1 MiB) / (365.25/2)(epochs in 1 year)
> ```
> And rent calculation is done with the f64 precision and the final result is truncated to u64 in lamports.

Or do it how the official `solana rent` command does it:
([source](https://github.com/solana-labs/solana/blob/3c5f505d3e3ded8b3110b44bc458d01701b93e4d/cli/src/cluster_query.rs#L2120-L2135))
```rs
    let epoch_schedule = rpc_client.get_epoch_schedule()?;
    let rent_account = rpc_client.get_account(&sysvar::rent::id())?;
    let rent: Rent = rent_account.deserialize_data()?;
    let rent_exempt_minimum_lamports = rent.minimum_balance(data_length);
    let seconds_per_tick = Duration::from_secs_f64(1.0 / clock::DEFAULT_TICKS_PER_SECOND as f64);
    let slots_per_year =
        timing::years_as_slots(1.0, &seconds_per_tick, clock::DEFAULT_TICKS_PER_SLOT);
    let slots_per_epoch = epoch_schedule.slots_per_epoch as f64;
    let years_per_epoch = slots_per_epoch / slots_per_year;
    let (lamports_per_epoch, _) = rent.due(0, data_length, years_per_epoch);
    let cli_rent_calculation = CliRentCalculation {
        lamports_per_byte_year: rent.lamports_per_byte_year,
        lamports_per_epoch,
        rent_exempt_minimum_lamports,
        use_lamports_unit,
    };
```

KartikSoneji:
@austbot 
IMO, most definitely yes.
The config rent is around 20 sol for a 10k collection, and even though you get it back, it is not trivial to liquidate that amount of money.
Paying week-to-week makes much more sense, especially if you know you are only going to be minting for a few days.

PR #897, opened 2021-11-04
Add max supply functionality in candy machine

jnuno98:
I have added a new feature in candy machine that allow the user to select a max supply for each NFT created when calling the upload function. 
This flag is optional and when not used, the NFT is created with a max supply of 0, as it is by default at this moment.

I have added a new argument in the upload command that receives the path to a json file that should have the following structure:
`{
    "0": 10,
    "1": 123,
    "2": 4
}`

This file matches the number of each NFT with the intended max supply.

I believe this is an important feature for people that want to have an option on the max supply of their NFTs.

### DIFF
--- a/js/packages/cli/src/candy-machine-cli.ts
+++ b/js/packages/cli/src/candy-machine-cli.ts
@@ -77,6 +77,10 @@ programCommand('upload')
     '-r, --rpc-url <string>',
     'custom rpc url since this is a heavy command',
   )
+  .option(
+    '-m, --max_supply <string>',
+    'max supply JSON file for each NFT (ex: {"0": 3, "1": 100})',
+  )
   .action(async (files: string[], options, cmd) => {
     const {
       number,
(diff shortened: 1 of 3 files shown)


js/packages/cli/src/candy-machine-cli.ts
```diff
@@ -90,6 +94,7 @@ programCommand('upload')
       retainAuthority,
       mutable,
       rpcUrl,
+      max_supply,
```
bhgames, line 97: camelCase please

js/packages/cli/src/commands/upload.ts
```diff
@@ -102,7 +105,7 @@ export async function upload(
             symbol: manifest.symbol,
             sellerFeeBasisPoints: manifest.seller_fee_basis_points,
             isMutable: mutable,
-            maxSupply: new BN(0),
+            maxSupply: new BN(max_supplies[i.toString()]),
```
bhgames, line 108: you realize this is only run once, right? Look at the if statement. i===0. So you can only set maxSupply once for the entire config - all share the same one. So there is no point in loading a file here - just take it as a single pass in option.

PR #914, opened 2021-11-06
fix: instruction chunks to optimize packs creation

m-sebastiyan:
When a card is added to a pack, UI sends two transactions: CreateTokenAccount and AddCardToPack. It's not optimal to have a big bunch of separate transactions due to fee charges, especially if a pack has a big number of cards. Also, in my case, I had transactions fail due to timeout. 

Steps to reproduce:
1. Set REACT_APP_ENABLE_NFT_PACKS=true in .env
2. Create 20 to 60 NFTs
3. Create NFT for voucher and mint some editions from it. 1 will be enough 
4. Go to http://localhost:3000/#/admin/pack/create/0
5. You will get a big number of sequential transactions that might get timeout since every card will have a separate transaction to call AddCardToPack and a separate transaction to create TokenAccount.

This fix makes transaction chunks for these instructions, to have fewer running transactions.
1. AddCardToPack instructions are grouped by 3 per transaction
2. CreateTokenAccount is grouped by 5 per transaction.


aheckmann (changes requested):
Code looks good. Please add instructions for how to test this or add automated unit tests which prove the fix works properly. We must be able to test the fix working before we can land it.

aheckmann:
Nice write up, thank you. Why do AddCardToPack and CreateTokenAccount use different batch sizes?

m-sebastiyan:
> AddCardToPack

AddCardToPack has a bigger size, I tried to set 4 or 3 for it, but then it would go over the limit of transaction size.

PR #921, opened 2021-11-08
Remove GraphQL cache server

lexich:


aheckmann:
Did you run through and test each package by hand to confirm the changes to yarn.lock don't break master?

lexich:
> Did you run through and test each package by hand to confirm the changes to yarn.lock don't break master?

Hi @aheckmann  This package was fully isolated and didn't expose anything outside. Also I've updated `yarn.lock` after removing this package.

PS: I think that we have GitHub actions for this purpose. Running test, lining, checking that `yarn.lock` is valid. But unfortunately `green tests` aren't strictly required for merging. E.g. right now we have `red tests` with linting because of code with linting errors
Screen Shot 2021-11-09 at 6 20 40 PM
.

aheckmann:
> This package was fully isolated and didn't expose anything outside. Also I've updated yarn.lock after removing this package.

Thank you

PR #939, opened 2021-11-10
Modify trade state account structure

RohanKapurDEV:
An example of how trade state accounts can be made more descriptive for better querying via memcmp filters from a client. There are a couple of good use cases for this. The first would be that it'd be simpler for an offchain crank to simply query the contract itself and isolate for the trade state accounts it needs to execute a sale. The other would be that if multiple markets all launched their own auction house contracts, having descriptive buyer/seller trade state accounts would allow people to track bids/asks on auction house accounts across all instances of the contract they're watching with a predictable interface made available via the account structure. 

The PR proposes such a change to the buyer trade state account. I only did it here as a template and if it looks good, will do the same thing for seller trade state accounts.

### DIFF
--- a/rust/auction-house/src/lib.rs
+++ b/rust/auction-house/src/lib.rs
@@ -1345,11 +1345,12 @@ pub const AUCTION_HOUSE_SIZE: usize = 8 + //key
 32 + //treasury mint
 32 + //authority
 32 + // creator
-8 + // bump
-8 + // treasury_bump
-8 + // fee_payer_bump
+1 + // bump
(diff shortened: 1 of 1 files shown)


bhgames:
Yes the problem is doing this makes it more expensive per offer or sale. The whole point of this is to take the information off chain to save on space and centralize it. The only part that remains on chain is the bare minimum to execute transactions. While it definitely would be nice to use memcmp to query faster, it would defeat the architectural choice made here which is to make creating an offer or a bid as cheap as humanly possible.

PR #940, opened 2021-11-10
FLP Fix: Wrap with try statements

bhgames:


js/packages/cli/src/fair-launch-cli.ts
```diff
@@ -2545,21 +2557,25 @@ program
...
+                  options: {
+                    commitment: 'single',
+                  },
+                  signers: [],
+                });
+              } catch (e) {
+                console.log('Skipping...');
+                console.error(e);
+              }
               console.log('Created...');
```
aheckmann, line 2579: think we need to move "Created.." up into the bottom of the `try`

PR #947, opened 2021-11-11
Adding support for a `secondary_token_mint`

jnwng:
the Candy Machine is currently designed to either accept lamports or SPL tokens. while this makes sense for conferring the cost of minting tokens out of the machine, in certain cases we may want to accept a _second_ token along with the mint cost. the primary use case is an "allowlist" — instead of checking wallets against addresses, we can instead use the ownership of a particular token as the "ticket" into the Candy Machine (sorta like what the Fair Launch Protocol does), along with the minting cost. an associated use case is something like "CAPTCHA", where a Captcha could be completed, a token is conferred, and the Captcha token + the minting cost can be passed on to the user. this approach is also a bit safer than allowlists that check wallet addresses, since once a "ticket" is transferred to a wallet, they can transfer to a burner for the eventual minting process (which is the general recommendation we give to people for safety!). additionally, if you wanted to get fancy you could charge two kinds of tokens (say, WOOF+SAMO) to get a doggo NFT. 

while it may have been simpler to configure the Candy Machine to accept only the combination of lamports + SPL tokens (since transferring lamports doesn't require a second token account), this is a more generic approach that will allow one to configure two SPL tokens for the Candy Machine. the CM can be initialized to accept Wrapped SOL in the case that they want to accept both SOL as well as the "ticket" (or any other token). 

i wanted to publish a version of this PR for review to validate directional alignment while i resolve some of the non-related issues, listed here:
- [x] it took a while, but i managed to get the Candy Machine running on localnet. i will fix the Anchor.toml accordingly
- [x] the CLI should be updated to be able to configure the secondary token price, and initialize the Candy Machine properly
- [x] it'd be nice to have a client-side example (perhaps a fork of exiled-apes/candy-machine-mint) that demonstrates a more accurate flow of data from the client. the tests here mix the "setup" with the actual invocation of the transactions, so it isn't the best example for users

open questions:
- should we throw a different error if there's either a mint mismatch or insufficient funds of the secondary token?
- i'm new to Solana programs, will throwing an "insufficient funds" error when checking the secondary token amount ensure that the tokens transferred for the first SPL token rewind automatically? i assume the instructions are all batched before execution
- ~~the way this is configured, it requires a `secondaryWallet` at initialization time for the Candy Machine. this is definitely a breaking change (the tests will fail for the SPL token / SOL usecase) — is there a way to make that an optional property via the Anchor macro?~~

[here's an example](https://secondary-token-mint.surge.sh/) of the `candy-machine-mint` application working with this Candy Machine ([code here](https://github.com/jnwng/candy-machine-mint/pull/1))


jnwng:
i was able to figure out how to (mostly) make the secondary wallet "optional" — i put it as a `remaining_account` (via https://github.com/project-serum/anchor/issues/1022). however, the `initializeCandyMachine` RPC call still requires a `secondaryPrice`, even if it set to null... this is probably okay since this RPC call is likely to come from the metaplex CLI itself. 

adding another task here to update the javascript CLI, too

jnwng:
i have added support for the secondary price into the CLI!

on devnet, the Candy Machine program id is [H8Le4ajukMFML7TvJciaFxnpYyTjK2CMpZ2jSaSCYJny](https://solscan.io/account/H8Le4ajukMFML7TvJciaFxnpYyTjK2CMpZ2jSaSCYJny?cluster=devnet). i uploaded a Candy Machine config [fnUC8MJ9YEMmdcB5fQetNRjbvNp2bDNjtT5HKdJ51wT](https://solscan.io/account/fnUC8MJ9YEMmdcB5fQetNRjbvNp2bDNjtT5HKdJ51wT?cluster=devnet) and created a Candy Machine @ [Bvg78t7wNJCkunm4YZRrJMmj9NqJDM28egeT9QpdNgQK](https://solscan.io/account/Bvg78t7wNJCkunm4YZRrJMmj9NqJDM28egeT9QpdNgQK?cluster=devnet)

here's a couple of the commands executing:
image

jnwng:
the tests also run against the deploy program in devnet via `anchor test --provider.cluster devnet`

jnwng:
i have a [working demonstration](https://secondary-token-mint.surge.sh/) of this program on devnet. had to bring up a different Candy Machine (config: `3gbv6xgEipQzNVfVexiVRJ4kQ3TR6mtnumasV5bCb6sv`, CM: `FL7sn6XAD8FuqAzmmPkSfDuYqaMQ3Vez5WXpRa4aBjHV`). requires wrapped SOL (`So11111111111111111111111111111111111111112`) and USDC (`8FRFC6MoGGkMFQwngccyu69VnYbzykGeez7ignHVAFSN`), the latter can be retrieved via this [generic faucet](https://spl-token-ui.netlify.app/#/token-faucets) — you will need your own ATA already set up (i will add a couple of faucet buttons on this page, i couldn't find two other generic tokens with faucets).

here's the NFT that came out of the Candy Machine! https://solscan.io/account/69xc4CrxW8pVCbM7Kf14fn3VTVbsiup86yPE2NHX5Agv

jnwng:
and the transaction information: https://solscan.io/tx/61dhuksZ93gC3yDfWtM9KhZnpVuuFQEYHWryTfSKziyM85LhC9cikVJ9kwMewXVHCrF5J9sp7LapoGko8RoWcEMe?cluster=devnet

rust/nft-candy-machine/src/lib.rs
```diff
@@ -76,6 +77,38 @@ pub mod nft_candy_machine {
...
+            if let Some(secondary_mint) = candy_machine.secondary_token_mint {
+                let secondary_wallet = &ctx.remaining_accounts[2];
+                let secondary_token_account_info = &ctx.remaining_accounts[3];
+                let transfer_authority_info = &ctx.remaining_accounts[1];
+                let secondary_token_account: spl_token::state::Account =
+                    assert_initialized(&secondary_token_account_info)?;
+
+                assert_owned_by(&secondary_token_account_info, &spl_token::id())?;
+                // Need to check that this destination wallet is owned by the program?
+                // assert_owned_by(&secondary_wallet, &spl_token::id())?;
```
jnwng, line 90: i am unsure of what to do here — i assume anchor is not validating these the way they do with the accounts in `context`. we're using `remaining_accounts` because this is an optional account to pass, so i _think_ i need to check that the parameterized `secondary_wallet` is... the wallet that we own?

jnwng:
alrighty @austbot im pretty happy with this — please let me know if there's a better way to test / demonstrate this working properly!

jnwng:
the original use-case here was to facilitate an allowlist distributed through the merkle-distributor but additionally being able to charge the user as well. in reviewing @lwus's work with gumdrop, it looks like there's a ton of overlap in functionality, and while gumdrop currently only charges sol / lamports, it skips the intermediary token that would be distributed that the Candy Machine would accept with this code. for the purposes of creating an allowlisted Candy Machine for presale, i think the Gumdrop approach is more "batteries included".

regarding the functionality introduced here, i had briefly discussed adding a secondary token mint to the Candy Machine with @bhgames and he raised a couple of suggestions around being able to issue CAPTCHA tokens to redeem (along with cost) against the Candy Machine. i'm wondering if there are other use-cases for a secondary token mint that are not overlapping with Gumdrop?

Issue #930, opened 2021-11-09
State 


aheckmann:
**Describe the bug**
There is a local UI state issue when switching from devnet to mainnet or changing stores which causes the app to not load properly.

Observations

1. Making changes to the packages/web/.env file and restarting the app does not always have the indented effect. The correct store or owner is not always used in RPC calls right away.
2. Switching from mainnet to devnet and vice versa does not always cause the app to connect to the configured environment, especially after changing the .env file and restarting

These smell like the same issue to me: a race condition in when we read state from the env at start up and connect to the correct environment vs reading the values from local storage or cookies etc.

The result is loss of precious time during development,.and a negative Developer Experience

Note: I've gotten into the habit of deleting the packages/web/.next directory but sometimes it's not enough. 

**To Reproduce**
1. Create a new wallet.
2. Start the app 
3. Connect the new wallet and set to devnet
4. Init a store / follow the instructions to update your .env file and restart the app
5. Create an NFT
6. Stop the app
7. `mv packages/web/.env packages/web/.env1`
8. Create a new wallet
1. Restart the app
1. Connect the new wallet and set to mainnet
2. Create a new store, init and create a new NFT
1. Stop the app
1. `mv packages/web/.env packages/web/.env2`
2.  `mv packages/web/.env1 packages/web/.env`
2. Restart the app

Continue similar combinations of these steps until you see the app not loading right, timing out, waiting for a long time.

**Expected behavior**
The app should always honor the settings in packages/web/.env and mainnet/devnet wallet setting

**Desktop (please complete the following information):**
 - OS: OSX
 - Browser: All

0xCryptoSheik:
@austbot @aheckmann is this no longer reproducible on your side after #958 ?

I expected #958 to be a good first step but not necessarily to fix this issue completely:
I believe there are still multiple areas in the code where a race condition or side-effect related to previous settings is posible.

Of course I'm also fine with creating more specific issues around this if/when needed.

PR #958, opened 2021-11-14
fix(connection): ensure stable network connection

0xCryptoSheik:
- changing the Solana "network" endpoint now reloads the app fully to avoid possible side-effects, race-conditions, bugs, ...
- the "network" endpoint is now determined only once on app mount, by priority:
  - from the `network` **hash-router querystring param**
  - from the `network` **local storage key/value**
  - **default** (`mainnet-beta` / `api.metaplex.solana.com`)
- `useQuerySearch` hook is now properly memoized to prevent unnecessary rendering of its consumers, if the `location` from react-router hash-router changes, but not the `search` property.
- `ConnectionContext` no longer triggers unnecessary re-renders of its consumers thanks to a memoized provider value.
- `ConnectionContext` now only exposes `{ connection: Connection, endpoint: Endpoint, tokens: Map<string, TokenInfo> }`, redundant values have been removed & migrated.
- general code-cleanup of packages/common/contexts/connection (types, naming, ...).
- miscellaneous changes (prettier, trivial types, ...)

Related to #930

### DIFF
--- a/js/packages/common/src/contexts/connection.tsx
+++ b/js/packages/common/src/contexts/connection.tsx
@@ -1,4 +1,10 @@
-import { sleep, useLocalStorageState } from '../utils/utils';
+import React, { useContext, useEffect, useMemo, useRef, useState } from 'react';
+import {
+  TokenInfo,
+  TokenListProvider,
+  ENV as ChainId,
+} from '@solana/spl-token-registry';
+import { WalletNotConnectedError } from '@solana/wallet-adapter-base';
 import {
   Keypair,
   clusterApiUrl,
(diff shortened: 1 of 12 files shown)


0xCryptoSheik:
cc @aheckmann: early draft, didn't test further than building `common` & `web` yet.

0xCryptoSheik:
Not sure I'm able to reproduce the issues described in #930 but this PR ensures the "network" is stable throughout the app's lifecycle, thus reduces the risk for this kind of side-effects / desync / race-conditions & allow to direct the bug hunt to other parts of the app. 

Tested, the new way of changing "network" works as intended, everything else works as well as before as far as I can tell.

Performance should be just a tiny bit better.

Ready for review.

js/packages/common/src/contexts/connection.tsx
```diff
@@ -13,171 +19,179 @@ import {
...
+  const { current: connection } = useRef(new Connection(endpoint.url));
 
-  const env =
-    ENDPOINTS.find(end => end.endpoint === endpoint)?.name || ENDPOINTS[0].name;
+  const [tokens, setTokens] = useState<Map<string, TokenInfo>>(new Map());
 
-  const [tokens, setTokens] = useState<TokenInfo[]>([]);
-  const [tokenMap, setTokenMap] = useState<Map<string, TokenInfo>>(new Map());
   useEffect(() => {
-    // fetch token files
```
0xCryptoSheik, line 113: On master branch, changing the "network" (or "env" as it's called there), while this token fetching is ongoing would cause undefined behavior.

aheckmann:
Thanks @0xCryptoSheik !

PR #959, opened 2021-11-14
Allow limited edition to inherit metadata mutability from master NFT.

### LINKED ISSUE #951: Allow mutable limited edition NFT metadata
**Is your feature request related to a problem? Please describe.**
I would like to be able to modify the metadata of limited edition NFTs to be different than their master edition's metadata, e.g. contain a different image uri and name.

**Describe the solution you'd like**
In the metadata program's source code, in `src/utils.rs` the `mint_limited_edition` function calls the `process_create_metadata_accounts_logic` function with a hard-coded `is_mutable = false` parameter, thus it's not possible to update the metadata of the newly minted "child edition" NFT.

**Describe alternatives you've considered**
I saw that NFT families on solanart consist of master edition NFTs whose metadata is straightforward to be set to a unique value. However, as far as I know, these master editions are not connected in any way, except for their metadata. It would be nice though, to keep the underlying logic/structure that master and subsequent editions have with each other.

**Additional context**
Maybe I'm missing something, or limited editions shouldn't be used like this, but it doesn't seem to be a huge task, and I would be happy to open a PR that enables updating limited edition NFT metadata, if it's something that would make sense.

PopcornPaws:
Aims to resolve #951

### DIFF
--- a/rust/token-metadata/program/src/utils.rs
+++ b/rust/token-metadata/program/src/utils.rs
@@ -592,7 +592,7 @@ pub fn mint_limited_edition<'a>(
         },
         master_metadata.data,
         true,
-        false,
+        master_metadata.is_mutable,
     )?;
     let edition_authority_seeds = &[
         PREFIX.as_bytes(),


bhgames:
It's part of the design that mutability is false for a limited edition. They should be static snapshots of their parent. We will not be merging this.

PR #965, opened 2021-11-15
chore: CI github actions cleanup

thlorenz:
First round of getting CI working again and improving granularity of tests.

- removed obsolete `pull-request.yml` which was all commented and caused CI errors on each push 
- cleaned up JS related workflows `lint.yml` and `yarn-and-test-cli.yml`
- added `build-and-test.yml` which tests rust programs via `cargo test-bpf`

### Build and Test Rust Programs

The following changes/additions were necessary to allow running `cargo test-bpf` in CI:

- install linux deps, see `.github/actions/install-linux-build-deps/action.yml` which install
  the dependencies needed to build + link bpf programs
- install solana, see `.github/actions/install-solana/action.yml ` which provides
  `cargo-build-bpf` and `cargo-test-bpf`
- install rust, see `.github/actions/install-rust/action.yml`

Caches are used to make solana and rust installs faster. Separate action files were created in
order to allow reuse across steps and possibly pull these out and publish at a later date.

Additionally `.github/actions/build-token-vault/action.yml` was included in order to allow
reuse across steps as the token-vault `.so` is needed by other programs, i.e. `token-metadata`.

The main workflow file `.github/workflows/build-and-test.yml` specifies `test-token-vault` and
`test-token-metatdata` as _needs_, which makes them execute separately and in parallel.
This is an improvement over just running `cargo test-*` on the entire repo for the following
reasons:

- tests run in isolation and in parallel
- when tests for a crate fail it is easily visible which it is via the github UI and doesn't
  require reading through the build logs
- cargo caches are isolated which allows caching the `target` dir (previously disabled due to
  conflicts between crates)

### Follow Up

The following tasks need to be completed in order to call this done:

- clean out the `ci` folder which is no longer needed most likely, i.e.
  `rust/ci/install-build-deps.sh` was converted to `.github/actions/install-solana/action.yml`
- cleanup unused code in token-metadata tests; need to decide if we want to keep the test
  utilities methods around, allowing _dead code_ or just remove them for now
- add more tests to the `token-metadata` program (at this point each instruction has about one
  _success_ tests and 2 failure cases)
- add CI tests for the remaining programs

## Checks

Since these workflows aren't on master _yet_ they don't run for _this_ PR that adds them.
However I ran them successully on my fork.

- [lint](https://github.com/thlorenz/metaplex/actions/runs/1463014487)
- [unit tests](https://github.com/thlorenz/metaplex/actions/runs/1463014481)
- [rust build and test](https://github.com/thlorenz/metaplex/actions/runs/1463014479)

## Facit

Given that CI is now intact and passing we should not merge any PRs that break it in the
future.


aheckmann:
> need to decide if we want to keep the test
utilities methods around, allowing dead code or just remove them for now

Commit history keeps a record of it so let's remove.

> Given that CI is now intact and passing we should not merge any PRs that break it in the
future.

👏👏👏

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,110 @@
...
+    paths:
+      - 'rust/**'
+  pull_request:
+    branches: [master]
+    paths:
+      - 'rust/**'
+
+env:
+  CARGO_TERM_COLOR: always
+  SOLANA_VERSION: 1.8.2
```
aheckmann, line 15: It would be nice to automatically test using the currently active versions of Solana on mainnet. I'm looking around for this info now

thlorenz: In that case we should create a reusable action to install `stable|beta` versions of solana.
We can look into that and update the workflow later, shouldn't block the PR though IMO to get CI fixed at least.

thlorenz:
> Commit history keeps a record of it so let's remove.

Agreed. Let's get this PR in and I can do a follow up PR which cleans up the two crates we cover with CI now.

Then I can work on CI for the remaining crates.

PR #966, opened 2021-11-15
Randomize upload

jjohnson5253:
@bhgames Here is a randomize function that people can use to randomize their NFT png/json pairs before uploading.
`ts-node metaplex/js/packages/cli/src/candy-machine-cli.ts randomize /path/to/NFTs`

I based it off this [stackoverflow post](https://stackoverflow.com/questions/2450954/how-to-randomize-shuffle-a-javascript-array?page=1&tab=votes#tab-top) which uses the Fisher-Yates shuffle algorithm. I modified it to change filenames.

I decided to shuffle the actual files instead of the cache content generated in the upload command for a couple reasons:
1. This would make sure the files are the same no matter how many times the user runs the upload cmd
2. The user knows which order the files are being uploaded in (they can see after they run the randomize cmd)


pijulius:
Not sure but shouldn't this also change the png filenames in the json file itself? as even if you rename 1.json to 0.json and then 1.png to 0.png the new 0.json file will still have a reference to 1.png:

"image": "1.png" and "properties": { "files": [{"url": "1.png" .....}]}

jjohnson5253:
I believe image and url properties of metadata get renamed by the upload script...

jjohnson5253:
looks like a shuffle implementation was made in cbb590d53a3a9718ecf2b526580952fc6d10baaf

PR #968, opened 2021-11-15
Fix race condition causing traits to be layered out of order

### LINKED ISSUE #967: Generative art is sometimes layered in the wrong order by the CLI tool
**Describe the bug**
There is a bug that results in trait PNG being layered in the wrong order in some cases. The portion of the CLI tool that does the layering is vulnerable to a race condition here: https://github.com/metaplex-foundation/metaplex/blob/master/js/packages/cli/src/commands/createArt.ts#L16. The promises are kicked off in the correct order but can finish in different orders. Most of the time, it works fine. However in some cases the promises finish in different orders.

**To Reproduce**
- Run the tool over and over again on the sample data with `-n <number_of_images>` set to at least 500
- Observe that in some cases the generated images are missing eyes, mouth, faces, etc. This happens when the layering is done out of order. For example, a face layered over the eyes hides the eyes completely.

**Expected behavior**
The generated image should ensure that each trait is rendered in the right order for each generated image.

**Screenshots**
Here is a good generated image:
correct-sample

Here are some bad generated images caused by this bug:
bad-sample-3
bad-sample-2
bad-sample-1

**Desktop (please complete the following information):**
 - OS: MacOS 11.6
 - Browser: Chrome
 - Version: 95.0.4638.69

candymandev:
Fixes #967

### DIFF
--- a/js/packages/cli/src/commands/createArt.ts
+++ b/js/packages/cli/src/commands/createArt.ts
@@ -13,15 +13,13 @@ function makeCreateImageWithCanvas(order, width, height) {
     return async function createImage(image) {
       const start = Date.now();
       const ID = parseInt(image.id, 10) - 1;
-      await Promise.all(
-        order.map(async cur => {
-          const imageLocation = `${TRAITS_DIRECTORY}/${cur}/${image[cur]}`;
-          const loadedImage = await loadImage(imageLocation);
-          context.patternQuality = 'best';
-          context.quality = 'best';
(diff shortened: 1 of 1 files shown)


bhgames (approved):

bhgames:
drop wallet for medal

candymandev:
@bhgames Awesome, here's my wallet: 2ZZkHXGqhXQFP9TiLkkA3hrebrsyJqGA2iUixwjsUjmK

Thanks!

PR #975, opened 2021-11-16
fixed two spinners

C0mberry:


austbot (approved):

C0mberry:
here i fixed two spinners on admin page and changed page rendering

Issue #980, opened 2021-11-16
Build Error: Property DesignAndInfo Does Not Exist on CreatePackSteps


statikdev:
**Describe the bug**
When building the repository after https://github.com/metaplex-foundation/metaplex/commit/d8202c30bda42b60fe6f4d8d08d34c1811250fa1, the following error is displayed:

```
$ tsc -p ./src
lerna ERR! yarn run build exited 1 in 'web'
lerna ERR! yarn run build stdout:
$ next build
info  - Loaded env from /Users/rob/Development/skulldiv/ogskulls-drop/js/packages/web/.env.production
info  - Loaded env from /Users/rob/Development/skulldiv/ogskulls-drop/js/packages/web/.env
info  - Using webpack 5. Reason: Enabled by default https://nextjs.org/docs/messages/webpack5
info  - Checking validity of types...
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run build stderr:
Failed to compile.

./src/views/packCreate/components/Header/data.ts:17:20
Type error: Property 'DesignAndInfo' does not exist on type 'typeof CreatePackSteps'.

  15 |   //   title: 'Pricing & Expiration ',
  16 |   // },
> 17 |   [CreatePackSteps.DesignAndInfo]: {
     |                    ^
  18 |     title: 'Design your pack',
  19 |   },
  20 |   [CreatePackSteps.ReviewAndMint]: {
error Command failed with exit code 1.

lerna ERR! yarn run build exited 1 in 'web'
lerna WARN complete Waiting for 2 child processes to exit. CTRL-C to exit immediately.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**To Reproduce**
1. download repository
2. yarn install && yarn bootstrap && yarn build

**Expected behavior**
The build should compile without error.

**Screenshots**

**Desktop (please complete the following information):**
- OS: Mac OS X

**Additional context**
directly related to the removal of the DesignAndInfo property in https://github.com/metaplex-foundation/metaplex/commit/d8202c30bda42b60fe6f4d8d08d34c1811250fa1

aheckmann:
Fixed in https://github.com/metaplex-foundation/metaplex/pull/989/files

PR #981, opened 2021-11-16
fix: comment out DesignAndInfo from Header per d8202c3

statikdev:
in commit d8202c3, the `DesignAndInfo` property reference was not removed from the `Header` component

fixes https://github.com/metaplex-foundation/metaplex/issues/980


aheckmann:
Fixed in https://github.com/metaplex-foundation/metaplex/pull/989/files

PR #984, opened 2021-11-16
Allow is_mutable to be flipped to false through update_metadata_accounts - WIP

samuelvanderwaal:
Currently there appears to be no way to set `is_mutable` to be `false` *after* an NFT metadata account has been created. Many NFTs, such as any created with default settings by candy_machine_cli, are created with the flag to `true`. It is desirable to have the ability to flip it to `false` at a later date, once any metadata issues are resolved or the community makes a decision to do so, etc.  

The code changes in this PR are relatively simple but require touching multiple programs in the repo (any program that calls `update_metadata_accounts` as it has an extra parameter now) so I'm not sure what the repo maintainers would like to see in terms of testing to allow this PR to be merged. 

An alternate approach would be to add a separate handler solely for flipping the `is_mutable` flag similar to the `process_update_primary_sale_happened_via_token` handler. This may be preferable for backwards compatibility and I would be happy to update the PR with this approach instead.


bhgames (changes requested):
I'm surprised this isnt' available already. I feel like I had it in there. In any case, this will require a new action. You are changing the signature of an action that may be used by many apps. We will need to see a new update endpoint that calls the same logic but has this extra field. Call it updateMetadataV2 or something.

samuelvanderwaal:
> I'm surprised this isnt' available already. I feel like I had it in there. In any case, this will require a new action. You are changing the signature of an action that may be used by many apps. We will need to see a new update endpoint that calls the same logic but has this extra field. Call it updateMetadataV2 or something.

Would it be better to just create a new handler entirely that only updates the mutability?
E.g. `set_metadata_immutable` or whatever. That seems to be a bit cleaner than having multiple versions of `update_metadata_accounts`.

samuelvanderwaal:
> @samuelvanderwaal I agree with @bhgames. A v2 instruction matches our previous way of updating.

Sounds good; will do.

austbot (approved):

samuelvanderwaal:
I've ported this PR over to the metaplex-program-library.

Issue #988, opened 2021-11-17
web CI is failing


aheckmann:
**Describe the bug**
CI is failing for the web directory

https://vercel.com/metaplex/metaplex-web/2PBqsasF9tuiAbTrgACXLHVubVE1

**To Reproduce**
Open a PR

**Expected behavior**
CI passes

**Additional context**

```


Detected Next.js version: 11.0.1
--
11:07:37.702 | Running "yarn run build"
11:07:38.053 | yarn run v1.22.17
11:07:38.092 | $ lerna run build
11:07:38.538 | lerna notice cli v3.22.1
11:07:38.541 | lerna info versioning independent
11:07:38.541 | lerna info ci enabled
11:07:38.608 | lerna info Executing command in 5 packages: "yarn run build"
11:08:11.863 | lerna info run Ran npm script 'build' in '@oyster/common' in 33.2s:
11:08:11.864 | $ run-s build-ts build-css
11:08:11.865 | $ tsc
11:08:11.865 | $ less-watch-compiler src/ dist/lib/ --run-once
11:08:11.865 | Running less-watch-compiler once.
11:08:26.775 | lerna info run Ran npm script 'build' in '@metaplex/cli' in 48.1s:
11:08:26.776 | $ tsc -p ./src
11:08:51.399 | lerna ERR! yarn run build exited 1 in 'web'
11:08:51.400 | lerna ERR! yarn run build stdout:
11:08:51.400 | $ next build
11:08:51.400 | info  - Loaded env from /vercel/path0/js/packages/web/.env.production
11:08:51.400 | info  - Loaded env from /vercel/path0/js/packages/web/.env
11:08:51.401 | info  - Using webpack 5. Reason: Enabled by default https://nextjs.org/docs/messages/webpack5
11:08:51.401 | info  - Checking validity of types...
11:08:51.401 | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
11:08:51.401 |  
11:08:51.401 | lerna ERR! yarn run build stderr:
11:08:51.401 | Failed to compile.
11:08:51.402 |  
11:08:51.402 | ./src/views/packCreate/components/Header/data.ts:17:20
11:08:51.402 | Type error: Property 'DesignAndInfo' does not exist on type 'typeof CreatePackSteps'.
11:08:51.402 |  
11:08:51.402 | [0m [90m 15 \| [39m  [90m//   title: 'Pricing & Expiration ',[39m[0m
[...]

PR #989, opened 2021-11-17
Fix metaplex-web CI

aheckmann:
Fixes the typescript issues causing CI to fail as reported in #988

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -13,6 +13,7 @@
   "scripts": {
     "bootstrap": "lerna link && lerna bootstrap",
     "build": "lerna run build",
+    "build-CI-without-gumdrop": "lerna run build --ignore gumdrop",
     "start": "cross-env CI=true lerna run start --scope @oyster/common --stream --parallel --scope web",
     "lint": "prettier -c 'packages/*/{src,test}/**/*.ts' && npm run lint:eslint",
     "lint:eslint": "eslint 'packages/*/{src,test}/**/*.ts'",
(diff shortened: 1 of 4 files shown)


js/package.json
```diff
@@ -13,6 +13,7 @@
   "scripts": {
     "bootstrap": "lerna link && lerna bootstrap",
     "build": "lerna run build",
+    "build-CI-without-gumdrop": "lerna run build --ignore gumdrop",
```
aheckmann, line 16: gumdrop causes Vercel to run out of memory. leaving out of CI until we move it all out of Vercel

js/packages/web/src/views/packCreate/components/Sidebar/data.ts
```diff
@@ -9,7 +9,7 @@ export const STEPS_TITLES: Record<CreatePackSteps, string> = {
   [CreatePackSteps.ReviewAndMint]: 'Review & Mint',
 };
 
-export const CONTINUE_TITLES: Record<CreatePackSteps, string> = {
+export const CONTINUE_TITLES: Record<number, string> = {
```
thlorenz, line 12: Not sure about this one. Why do we need to generalize the type to a number when it is really more specifically a particular enum?

aheckmann: enums values (the way we defined the enum here) are numbers. typescript fails without this.

aheckmann: the enum is the same as `{ first: 0, second: 1, third: 2 }`, hence the number type

thlorenz: OK it's fine for now. We should probably use `StringUnion`s anyways which are basically a superset of enums.

austbot (approved):

PR #1003, opened 2021-11-18
fixed build

C0mberry:
Commented out a line where was import to commented enum item. Build didn't work because this

### DIFF
--- a/js/packages/web/src/views/packCreate/components/Header/data.ts
+++ b/js/packages/web/src/views/packCreate/components/Header/data.ts
@@ -11,12 +11,6 @@ export const HEADER_CONTENT: Record<CreatePackSteps, HeaderContentRecord> = {
   [CreatePackSteps.AdjustQuantities]: {
     title: 'Adjust Quantities',
   },
-  // [CreatePackSteps.SalesSettings]: {
-  //   title: 'Pricing & Expiration ',
-  // },
-  // [CreatePackSteps.DesignAndInfo]: {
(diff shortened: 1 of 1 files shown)


m-sebastiyan (approved):

js/packages/web/src/views/packCreate/components/Header/data.ts
```diff
@@ -14,9 +14,9 @@ export const HEADER_CONTENT: Record<CreatePackSteps, HeaderContentRecord> = {
   // [CreatePackSteps.SalesSettings]: {
   //   title: 'Pricing & Expiration ',
   // },
-  [CreatePackSteps.DesignAndInfo]: {
-    title: 'Design your pack',
-  },
+  // [CreatePackSteps.DesignAndInfo]: {
+  //   title: 'Design your pack',
+  // },
```
austbot, line 19: @m-sebastiyan  and @C0mberry  instead of commenting this out lets put it behind the feature flag.

austbot (changes requested):

js/packages/web/src/views/packCreate/components/Header/data.ts
m-sebastiyan: I would also delete other stuff that's been commented out here https://github.com/metaplex-foundation/metaplex/commit/d8202c30bda42b60fe6f4d8d08d34c1811250fa1

austbot (approved):

PR #1005, opened 2021-11-18
removed redundant env variable

C0mberry:
removed useless variable from env and front


js/packages/web/.env
```diff
@@ -1,5 +1,4 @@
 REACT_APP_STORE_OWNER_ADDRESS_ADDRESS=
-REACT_APP_STORE_ADDRESS=
```
austbot, line 2: this is still needed to have a slightly quicker store load

austbot (changes requested):

PR #759, opened 2021-10-19
GIF support for Candy Machine

lettuce0x:
Add support to upload/verify gif images. It fixes #511

   - Update the upload command to accept --image-type gif|png (default: png)
   - Update the verify_token_metadata command to accept --image-type gif|png (default: png)
   - Update arweave and aws uploaders to accept gif
   
   
   Command example using gif images:
   
   ```bash
   metaplex upload ~/nfts/gifs --image-type gif --keypair ~/.config/solana/id.json
   metaplex verify_token_metadata ~/nfts/gifs --image-type gif --keypair ~/.config/solana/id.json
   ```
   
   If you don't add the image-type argument, it will use png, the same behavior before this PR.

PR #689, opened 2021-10-10
add glb extension and rename .png by EXTENSION_PNG

Exaustia:
### Feature

- Change .png for EXTENSION_PNG in the files `js/packages/cli/src/commands/upload.ts` and `js/packages/cli/src/helpers/upload/arweave.ts`
- Add EXTENTION_GLB in `js/packages/cli/src/helpers/constants.ts`
- Add GLB type in the process of the Metadata creation

PR #1028, opened 2021-11-20
Added support for GIF to the candy-machine-cli upload command

Tishj:
I needed it for my project, so I figured I would investigate why the image had to be PNG to work with the candy-machine-cli script
turns out it doesn't have to be, it was just hardcoded to only take png, so I fixed that.

### DIFF
--- a/js/packages/cli/src/candy-machine-cli.ts
+++ b/js/packages/cli/src/candy-machine-cli.ts
@@ -19,6 +19,7 @@ import {
   CONFIG_LINE_SIZE,
   EXTENSION_JSON,
   EXTENSION_PNG,
+  EXTENSION_GIF,
   CANDY_MACHINE_PROGRAM_ID,
 } from './helpers/constants';
 import {
@@ -118,29 +119,29 @@ programCommand('upload')
       secretKey: ipfsInfuraSecret,
     };
 
-    const pngFileCount = files.filter(it => {
-      return it.endsWith(EXTENSION_PNG);
+    const imageCount = files.filter(it => {
+      return it.endsWith(EXTENSION_PNG) || it.endsWith(EXTENSION_GIF);
     }).length;
     const jsonFileCount = files.filter(it => {
       return it.endsWith(EXTENSION_JSON);
     }).length;
 
     const parsedNumber = parseInt(number);
-    const elemCount = parsedNumber ? parsedNumber : pngFileCount;
+    const elemCount = parsedNumber ? parsedNumber : imageCount;
 
-    if (pngFileCount !== jsonFileCount) {
+    if (imageCount !== jsonFileCount) {
       throw new Error(
-        `number of png files (${pngFileCount}) is different than the number of json files (${jsonFileCount})`,
+        `number of png files (${imageCount}) is different than the number of json files (${jsonFileCount})`,
       );
     }
 
-    if (elemCount < pngFileCount) {
+    if (elemCount < imageCount) {
       throw new Error(
-        `max number (${elemCount})cannot be smaller than the number of elements in the source folder (${pngFileCount})`,
+        `max number (${elemCount})cannot be smaller than the number of elements in the source folder (${imageCount})`,
       );
     }
 
-    log.info(`Beginning the upload for ${elemCount} (png+json) pairs`);
+    log.info(`Beginning the upload for ${elemCount} (image+json) pairs`);
 
     const startMs = Date.now();
     log.info('started at: ' + startMs.toString());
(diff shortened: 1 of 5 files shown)


js/packages/cli/src/candy-machine-cli.ts
```diff
@@ -118,29 +119,29 @@ programCommand('upload')
...
 
     const parsedNumber = parseInt(number);
-    const elemCount = parsedNumber ? parsedNumber : pngFileCount;
+    const elemCount = parsedNumber ? parsedNumber : imageCount;
 
-    if (pngFileCount !== jsonFileCount) {
+    if (imageCount !== jsonFileCount) {
       throw new Error(
-        `number of png files (${pngFileCount}) is different than the number of json files (${jsonFileCount})`,
+        `number of png files (${imageCount}) is different than the number of json files (${jsonFileCount})`,
```
austbot, line 134:
```suggestion
        `number of image files (${imageCount}) is different than the number of json files (${jsonFileCount})`,
```

js/packages/cli/src/candy-machine-cli.ts
```diff
@@ -118,29 +119,29 @@ programCommand('upload')
...
       );
     }
 
-    if (elemCount < pngFileCount) {
+    if (elemCount < imageCount) {
       throw new Error(
-        `max number (${elemCount})cannot be smaller than the number of elements in the source folder (${pngFileCount})`,
+        `max number (${elemCount})cannot be smaller than the number of elements in the source folder (${imageCount})`,
```
austbot, line 140:
```suggestion
        `max number (${elemCount}) cannot be smaller than the number of elements in the source folder (${imageCount})`,
```

js/packages/cli/src/helpers/cache.ts
```diff
@@ -29,6 +31,7 @@ export function saveCache(
 ) {
   cacheContent.env = env;
   cacheContent.cacheName = cacheName;
+//   console.log(`SAVING CACHE: ${JSON.stringify(cacheContent)}`);
```
austbot, line 34: please remove commented code

austbot (changes requested):
Please run lint and fix comments, and then it looks good to go

thlorenz:
The following PRs are related:

-  https://github.com/metaplex-foundation/metaplex/pull/759
-  https://github.com/metaplex-foundation/metaplex/pull/689

Do we need a more general solution here which supports a variety of image formats and/or should we consolidate the changes made into one solution covering `png`, `glb` and `gif` for now? cc  @austbot

thlorenz:
Closing in favor of a more [generic solution](https://github.com/metaplex-foundation/metaplex/issues/1044).
Please provide a PR more along those lines and/or update this one here.

Tishj:
That's entirely fair
I was attempting to make it more generic with the switch case, and grabbing the extension from the file itself
but what I have written here is very limited, I haven't updated any tests or even the `verify` option. Nor have I written upload support for these other types in any other upload provider than `arweave`

Furthermore there needs to be an additional change in the script, I couldn't get arweave to detect the link between the 'image' and the `uri` portion of the metadata
I had to change my generator to change `.png` to `.gif` beforehand to get this to work properly

redhatt06:
> I needed it for my project, so I figured I would investigate why the image had to be PNG to work with the candy-machine-cli script turns out it doesn't have to be, it was just hardcoded to only take png, so I fixed that.

Hi

> That's entirely fair I was attempting to make it more generic with the switch case, and grabbing the extension from the file itself but what I have written here is very limited, I haven't updated any tests or even the `verify` option. Nor have I written upload support for these other types in any other upload provider than `arweave`
> 
> Furthermore there needs to be an additional change in the script, I couldn't get arweave to detect the link between the 'image' and the `uri` portion of the metadata I had to change my generator to change `.png` to `.gif` beforehand to get this to work properly

Hi Tishj,

I have been trying to upload a gif as well. After seeing your post I also have configured the hard coded part to accept gifs. It worked. However, after minting the gifs I am unable to see the gif. In the wallet the nft is just a loading animation and when I display it through solscan it leads me to a not found page. The problem seems to be related to my json file, I do not know how the json file should be configured for gifs. Could you help me out if you know what to do about it? 
ask

Tishj:
@redhatt06 yes of course!

So I also faced your issue, what's happening is that the image and the files[0].uri aren't being changed to the arweave url

I tried to fix it by changing the script, but was getting headaches
What fixed it for me was changing the 'png' to 'gif' in the asset json (0.json for example) if the associated image was a gif
(Before running 'upload')

ahl1395:
How did you guys change the hardcoded part where it only accepts PNGs? @redhatt06 @Tishj

PR #791, opened 2021-10-24
Align lifetime of process_instruction with solana

Gerhut:
Fixes #446
Ref: <https://docs.rs/solana-program/1.8.1/solana_program/entrypoint/type.ProcessInstruction.html>

PR #1039, opened 2021-11-23
Align lifetime of process_instruction with solana

### LINKED ISSUE #446: Cannot use `spl_token_metadata` with `solana_program_test`
Hello,
I am trying to combine a basic example of `solana_program_test` utility with the `spl_token_metadata` program, like so:

```rust
use solana_program_test::{tokio, processor, ProgramTest},

#[tokio::test]
async fn my_test() {
    // ...
    let mut program = ProgramTest::default();
    program.add_program(
        "spl_token_metadata",
        spl_token_metadata::ID,
        processor!(spl_token_metadata::processor::process_instruction),
    );
    // ...
}
```

But because of constraints enforced on lifetimes by the metadata entry point ([see here](https://github.com/metaplex-foundation/metaplex/blob/master/rust/token-metadata/program/src/processor.rs#L40)), I get the following error:

```
error[E0308]: mismatched types
   --> tests/my_test.rs:269:20
    |
269 |         processor!(spl_token_metadata::processor::process_instruction),
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ one type is more general than the other
    |
    = note: expected fn pointer `for<'r, 's, 't0, 't1> fn(&'r Pubkey, &'s [AccountInfo<'t0>], &'t1 [u8]) -> Result<_, _>`
               found fn pointer `for<'a, 'r> fn(&'a Pubkey, &'a [AccountInfo<'a>], &'r [u8]) -> Result<_, _>`
```

I tried many things such as wrapping the entrypoint or expanding the `processor!` macro but could not change the core error. :/
Anybody is using `solana_program_test`? How do you do it?

Gerhut:
Continues #791 sorry for my rust analyzer was not working so that I thought it passes the compile.

I removed and split all required lifetimes and it compiles but still not pass the test: `fair_launch` is ambiguous (name vs any other name during import resolution) which is the same as the master branch. (https://github.com/project-serum/anchor/issues/109 rename the program name resolves)

Fixes #446
Ref: https://docs.rs/solana-program/1.8.1/solana_program/entrypoint/type.ProcessInstruction.html


Gerhut:
kindly ping @thlorenz since I may do not know about how to test

Gerhut:
@austbot Just retried the compiling and it's success. And I can seem the 3 rust CI jobs are passed.

Could you share me the compile error message on your side?

thlorenz:
WRT to compiling in CI keep in mind that only the `token-vault` + `token-metadata` programs and dependents are currently compiled. We're working on covering more crates in CI. So it is possible that some crates are still not compiling properly like @austbot mentioned.

Gerhut:
I ran `cargo build` and `cargo build-bpf` in the `/rust` directory that seems compile all projects in the rust workspace.

PR #1054, opened 2021-11-24
gh actions: always run required workflows

aheckmann:
Otherwise pull requests become stuck when the workflow does not run but is
configured as required.

### DIFF
--- a/.github/workflows/yarn-and-test-cli.yml
+++ b/.github/workflows/yarn-and-test-cli.yml
@@ -1,14 +1,10 @@
-name: Yarn Install and Test JS
+name: Yarn Install and Test CandyMachine
 
 on:
   push:
     branches: [master]
-    paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/yarn-and-test-cli.yml
```diff
@@ -1,14 +1,10 @@
-name: Yarn Install and Test JS
+name: CandyMachine: Yarn Install and Test JS
```
thlorenz, line 1: This is not specific to CandyMachine, why the name change?

thlorenz (changes requested):
LGTM except for that naming to candy machine thingy.

.github/workflows/yarn-and-test-cli.yml
aheckmann: cli is the candy machine

PR #1069, opened 2021-11-29
github: add build workflow

aheckmann:
Triggers `yarn install` and `build` from the `js` directory which triggers a build in all packages.

This replaces the current [Vercel metaplex-web CI step](https://vercel.com/metaplex/metaplex-web) which runs out of memory on Vercel.

With this new workflow we can remove the `build-CI-without-gumdrop` package.json script currently used by the Vercel job.

### DIFF
--- a/.github/workflows/yarn-build.yml
+++ b/.github/workflows/yarn-build.yml
@@ -1,0 +1,26 @@
+name: Yarn build
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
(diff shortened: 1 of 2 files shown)


aheckmann:
The Vercel web check failed due to this PR removing the package script that skips gumdrop builds. Because gumdrop is not building, let's hold off on landing this PR until that is fixed first.

aheckmann:
Fixed Vercel CI. This is ready.

austbot (approved):

PR #1075, opened 2021-11-30
fix: Prevent number of hooks from changing in user badge component

steveluscher:
The number and call order of hooks in React components has to be the same from render to render. We had a situation in the user badge component where we'd return early before calling the `useTokenList()` hook.

That led to this error:

```
Warning: React has detected a change in the order of Hooks called by CurrentUserBadge. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://reactjs.org/link/rules-of-hooks
```

This PR fixes that. We might also consider turning on the React hooks linter (add `'plugin:react-hooks/recommended'` to the eslintrc) which would have warned about this like so:

image

Before we consider doing that, we should probably clean up the rest of the lint that's snuck into React components. See #1076.

### DIFF
--- a/js/packages/web/src/components/CurrentUserBadge/index.tsx
+++ b/js/packages/web/src/components/CurrentUserBadge/index.tsx
@@ -209,13 +209,14 @@ export const CurrentUserBadge = (props: {
   const { account } = useNativeAccount();
   const solPrice = useSolPrice();
   const [showAddFundsModal, setShowAddFundsModal] = useState<Boolean>(false);
+  const tokenList = useTokenList();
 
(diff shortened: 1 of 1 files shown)


steveluscher:
Does commenting on this PR reopen it?

steveluscher:
Retargeted to `master`.

steveluscher:
Retargeted to `master`, @roederw!

stegaBOB:
Reviewing this PR and hopefully merging it in tomorrow! Sorry for letting this sit for so long!

stegaBOB (approved):
Super quick fix to review! Looks good to me!

PR #1077, opened 2021-11-30
fix: Prevent wallet errors from bubbling up to the default error handler

steveluscher:
Before, if you rejected a wallet authorization in dev mode you would see the error bubble all the way up to Next's error handler.

bad

After this PR, you and your users both see the toast, but you, the dev, otherwise don't see a redbox.

good


roederw (approved):

roederw:
Thanks for the fix!

PR #1078, opened 2021-11-30
Fix expectedImagePath to be the same as the index of the metadata file

febo:
- Fixed the 'verify_token_metadata' command in the candy machine to use the index of the metadata file to check the image path ('0.json' -> '0.png') - currently is expecting all images to be name 'image.png'.
- Changed the way the timestamp of the execution is printed on the console


js/packages/cli/src/commands/verifyTokenMetadata/index.ts
```diff
@@ -90,7 +90,9 @@ export const verifyCreatorCollation = (
 };
 
 export const verifyImageURL = (image, files, manifestFile) => {
-  const expectedImagePath = `image${EXTENSION_PNG}`;
+  // The image is expected to have the same name as the index
+  const fileIndex = manifestFile.split('/').pop().split('.')[0];
```
C0mberry, line 94: you can pass tokenMetadata to this function from verifyMetadataManifests and use tokenMetadata.image

b2kdaman (approved):

austbot (approved):

js/packages/cli/src/candy-machine-cli.ts
```diff
@@ -291,14 +291,19 @@ programCommand('verify_token_metadata')
     const { number } = cmd.opts();
 
     const startMs = Date.now();
-    log.info('started at: ' + startMs.toString());
+    log.info(
```
thlorenz, line 350: This does quite a bit of work to generate the log message and should be inside something like 
```
if (log.info.enabled) {
 ...
}
```
To avoid doing that when we're not logging.

js/packages/cli/src/candy-machine-cli.ts
```diff
@@ -291,14 +291,19 @@ programCommand('verify_token_metadata')
...
+    );
     verifyTokenMetadata({ files, uploadElementsCount: number });
 
     const endMs = Date.now();
     const timeTaken = new Date(endMs - startMs).toISOString().substr(11, 8);
     log.info(
-      `ended at: ${new Date(endMs).toString()}. time taken: ${timeTaken}`,
+      `==> Verification ended: ${new Date(endMs).toString().split(' G')[0]}`,
     );
+    log.info(`Elapsed time: ${timeTaken}\n`);
```
thlorenz, line 362: Ditto above .. also please don't duplicate the date _prettyfing_ code pull out into a separate function or use something like [date-time](https://github.com/sindresorhus/date-time)

js/packages/cli/src/commands/verifyTokenMetadata/index.ts
thlorenz, line 93: Wondering if using [url](https://nodejs.org/api/url.html) or similar would do a better job at pulling out parts of this url.

thlorenz (changes requested):

austbot (approved):

PR #1086, opened 2021-12-01
enhancement: Fetch the token list once on startup; save folks 200K

steveluscher:
Every time that you call `TokenListProvider.resolve()` a new network fetch for the token list is made. We've been fetching two identical copies of the token list on startup!

In this PR, we create a singleton that everyone can share. The first consumer causes the token list to be fetched, and then every consumer after uses the cached result. This saves folks from having to download about 200K on startup.

Before:

image

After:

image

### DIFF
--- a/js/packages/web/src/contexts/tokenList.tsx
+++ b/js/packages/web/src/contexts/tokenList.tsx
@@ -34,7 +34,7 @@ export function SPLTokenListProvider({ children = null as any }) {
     ]: [WRAPPED_SOL_MINT]
 
   useEffect(() => {
-    getTokenListContainerPromise().then(setTokenList);
+    getTokenListContainerPromise().then(()=>setTokenList);
   }, [setTokenList]);
 
   const hasOtherTokens = !!process.env.NEXT_SPL_TOKEN_MINTS;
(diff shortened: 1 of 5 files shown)


steveluscher:
Retargeted to `master` @roederw!

stegaBOB:
Thank you for the PR! Planning to take a closer look and do some testing tomorrow! Hopefully can get it merged in by end of weekend at the latest.

stegaBOB (dismissed):
Looks good to me. Seems like there are some build issues with the vercel deploy though :(

stegaBOB (approved):
take 2?

js/packages/web/src/contexts/tokenList.tsx
```diff
@@ -34,7 +34,7 @@ export function SPLTokenListProvider({ children = null as any }) {
     ]: [WRAPPED_SOL_MINT]
 
   useEffect(() => {
-    getTokenListContainerPromise().then(setTokenList);
+    getTokenListContainerPromise().then(()=>setTokenList);
```
steveluscher, line 37: Hmm… the old code used to _run_ the `setTokenList` function, but the new code just _returns_ it without running. I’m pretty sure this means the token list is no longer being set.

I haven’t looked yet, what was the build error?

PR #1094, opened 2021-12-02
Adds a store v2 init implementatioin that allows a uri to point to an off chain settings resource

austbot:
For no code, we need a way to allow a settings URL to be represented on-chain to allow a no-code serverless store to have some settings and user preference state


rust/metaplex/program/tests/create_store.rs
```diff
@@ -0,0 +1,67 @@
...
+        let store = context.banks_client.get_account(store_key).await.unwrap().unwrap();
+        let store_data: Store = try_from_slice_unchecked(&store.data).unwrap();
+        let config = context.banks_client.get_account(config_key).await.unwrap().unwrap();
+        let config_data: StoreConfig = try_from_slice_unchecked(&config.data).unwrap();
+       
+        assert_eq!(store_data.key, Key::StoreV1);
+        assert_eq!(config_data.key, Key::StoreConfigV1);
+        assert_eq!(config_data.settings_uri, Some("https://notgoogle.com".to_string()));
+    }
+}
```
thlorenz, line 67: Ideally we'd add some more test here to check the _not happy_ paths.
In particular we should check for known attack vectors and at the very least cover the case where the `store_config` isn't empty but not owned by the program.

thlorenz (changes requested):
Did add some suggestions myself and would recommend to have someone more versed in contract security review this as well as I may have missed something in that regard.

rust/metaplex/program/src/processor/set_store.rs
```diff
@@ -88,3 +86,108 @@ pub fn process_set_store<'a>(
...
+        admin_wallet_info,
+        payer_info,
+    );
+    if res.is_err() {
+        return res;
+    }
+    if !store_config_info.data_is_empty() {
+        assert_owned_by(store_config_info, program_id)?;
+    }
+    let store_config_bump = assert_derivation(
```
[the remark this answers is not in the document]
austbot: I dont agree, this would open a security hole where someone could pass in a valid and init config of a different store, and just because its owned by the program does not mean its derived from the users wallet seed.

rust/metaplex/program/tests/create_store.rs
austbot: Sure, but its better than what we have which is no tests.

.github/workflows/build-and-test.yml
```diff
@@ -107,4 +108,39 @@ jobs:
           cargo +${{ env.RUST_TOOLCHAIN }} test -- --nocapture --test-threads 1
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf -- --nocapture --test-threads 1
+  test-metaplex-program:
```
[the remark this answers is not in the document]
austbot: okay will change

rust/metaplex/program/tests/create_store.rs
thlorenz: Agreed, however you're adding a new contract and we don't just need to make sure that it works, but also that it is secure and handles incorrect inputs correctly.
That is my opinion at least.

rust/metaplex/program/src/processor/set_store.rs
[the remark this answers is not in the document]
thlorenz: Not sure what you mean, the logic is exactly the same unless I'm missing something.
It's just rewritten to be easier to follow (at least for me).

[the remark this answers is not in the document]
thlorenz: I understand now that `assert_derivation` does two things and needs to run at all times.

rust/metaplex/program/src/instruction.rs
```diff
@@ -1099,6 +1120,40 @@ pub fn create_set_store_instruction(
...
+    admin: Pubkey,
+    payer: Pubkey,
+    public: bool,
+    settings_uri: Option<String>,
+) -> Instruction {
+    let accounts = vec![
+        AccountMeta::new(store, false),
+        AccountMeta::new(config, false),
+        AccountMeta::new_readonly(admin, true),
+        AccountMeta::new_readonly(payer, true),
```
VadimGrozinok, line 1137: payer should be writable because this account will pay for new account creation
so here should be AccountMeta::new(payer, true)

rust/metaplex/program/src/processor/set_store.rs
```diff
@@ -88,3 +86,108 @@ pub fn process_set_store<'a>(
...
+        auction_program_info,
+        token_vault_program_info,
+        rent_info,
+        system_info,
+        token_metadata_program_info,
+        token_program_info,
+        store_info,
+        admin_wallet_info,
+        payer_info,
+    );
```
VadimGrozinok, line 153: it would be better to use " ? " operator here so we don't need check result with " if "

aheckmann (approved):

aheckmann (approved):
+1 pending Thorsten's suggestions

thlorenz (approved):

PR #1106, opened 2021-12-03
Add gatekeeper to candy machine

Buzzec:


rust/nft-candy-machine-v2/src/lib.rs
```diff
@@ -59,6 +62,34 @@ pub mod nft_candy_machine_v2 {
...
+                remaining_accounts_counter += 1;
+                solana_gateway::Gateway::verify_and_expire_token(
+                    gateway_token.clone(),
+                    wallet.deref().clone(),
+                    &gatekeeper.gatekeeper_network,
+                    network_expire_feature.clone(),
+                    None,
+                )?;
+            } else {
+                solana_gateway::Gateway::verify_gateway_token_account_info(
```
austbot, line 85: I don't see an import, so its not compiling.

austbot (changes requested):

PR #1173, opened 2021-12-09
show funds modal issue when buying sale without enough balance

yamijuan:
- Refresh button loads token accounts with dropdown for all metadata
- show funds modal issue when buying sale without enough balance

image

image


js/packages/web/src/components/AuctionCard/index.tsx
```diff
@@ -355,6 +360,12 @@ export const AuctionCard = ({
     setLoading(false);
   };
   const instantSaleAction = () => {
+
+    if (balance.balanceLamports < (instantSalePrice?.toNumber()  || 0)) {
```
zaxozhu, line 364:
```suggestion
    const isNotEnoughLamports = balance.balanceLamports < instantSalePrice
    if (isNotEnoughLamports) {
```

js/packages/web/src/components/AuctionCard/index.tsx
```diff
@@ -355,6 +360,12 @@ export const AuctionCard = ({
...
+      setShowFundsIssueModal(true);
```
zaxozhu, line 365:
```suggestion
      return setShowFundsIssueModal(true);
```

js/packages/web/src/components/AuctionCard/index.tsx
```diff
@@ -251,6 +253,9 @@ export const AuctionCard = ({
 
   //console.log("[--P]AuctionCard", tokenInfo, mintKey)
   const myPayingAccount = balance.accounts[0];
+  const instantSalePrice = useMemo(() =>
+    auctionView.auctionDataExtended?.info.instantSalePrice
```
zaxozhu, line 257:
```suggestion
    auctionView.auctionDataExtended?.info.instantSalePrice.toNumber() || 0,
```

js/packages/web/src/components/AuctionCard/index.tsx
```diff
@@ -251,6 +253,9 @@ export const AuctionCard = ({
...
+    , [auctionView.auctionDataExtended]);
```
zaxozhu, line 258:
```suggestion
   [auctionView.auctionDataExtended]);
```

zaxozhu (changes requested):

austbot (approved):

austbot (changes requested):
please solve @zaxozhu  suggestions

js/packages/web/src/components/AuctionCard/index.tsx
yamijuan: can't do that. That variable is used in a transaction

js/packages/web/src/components/AuctionCard/index.tsx
yamijuan: same reason as above

js/packages/web/src/components/AuctionCard/index.tsx
yamijuan: this does not return anything and I believe it is more semantic to leave the return empty.

zaxozhu (approved):

Issue #1190, opened 2021-12-11
upload is expecting ".json.json" for the json files.


numisman:
**Describe the bug**
When you use the current master branch (#1172) and attempt to upload using the candy machine cli, you will encounter an error where the job is expecting a file with an extension of ".json.json"

**To Reproduce**
1. clone the current master
2. attempt to use the upload CLI
``` npx ts-node $CM_HOME/candy-machine-cli.js upload output --env $ENV --keypair $WALLET --log-level $LOGLEVEL --cache-name $CACHENAME --rpc-url $RPC_URL```
3. You will get an error 
``` upload was not successful, please re-run.  Error: ENOENT: no such file or directory, open 'assets\0.json.json'```

**Expected behavior**
.png and .json are uploaded to arweave successfully.

**Screenshots**
![ss_json_json](https://user-images.githubusercontent.com/87335441/145694094-a410f72b-f83c-46db-8411-c4d91ff672a0.png)

![ss_codeToFix](https://user-images.githubusercontent.com/87335441/145694120-02b660fe-1187-43d7-940d-7d6d09e6049c.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Brave

numisman:
The code screenshot is from metaplex/js/packages/cli/build/commands/upload.ts and shows where I think the change needs to be made.

brantchoate:
same problem. when I remove the extra ".json" I get a timeout. Not sure if it's related or another problem I'm having. All of my other config was working before upgrading though

anzeskulj02:
error can be fixed by changing 366 line in file js/packages/cli/src/commands/upload.ts to `assetKey.index,` . Error is now fixed

stegaBOB:
Just merged in the fix. This issue should be resolved

PR #1194, opened 2021-12-12
added fix for json.json error

amarjeet987:
Fix for https://github.com/metaplex-foundation/metaplex/issues/1190

```
ts-node js/packages/cli/src/candy-machine-cli.ts upload ./assets --env devnet --storage arweave --keypair ~/.config/solana/devnet-test.json
```

This command was causing issues for some json files and was throwing the following error : 
```
upload was not successful, please re-run. Error: ENOENT: no such file or directory, open 'assets\0.json.json'
```
I tried removing the hardcoded json as a whole from the upload.ts file, then it was working. But then again that might break something that was working initially with the hardcoded value, so kept the fix conditional depending upon the file name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ docs/lockup-ui/
 .DS_Store
 *~
 .idea
-assets
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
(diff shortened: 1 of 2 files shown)


stegaBOB (approved):

dommj00:
Hi, I'm kinda new to this, how do i apply the json json fix? Thanks

PR #1200, opened 2021-12-13
fix: fixed uploading step

ikadushkina:
Task: https://app.asana.com/0/1201289799176936/1201485500072261
fixed removing image by click on trash icon
fixed uploading image by url

### DIFF
--- a/js/packages/web/src/views/artCreate/index.tsx
+++ b/js/packages/web/src/views/artCreate/index.tsx
@@ -328,7 +328,7 @@ const UploadStep = (props: {
 
   const [customURL, setCustomURL] = useState<string>('');
   const [customURLErr, setCustomURLErr] = useState<string>('');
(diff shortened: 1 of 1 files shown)


js/packages/web/src/views/artCreate/index.tsx
```diff
@@ -398,6 +398,10 @@ const UploadStep = (props: {
           accept=".png,.jpg,.gif,.mp4,.svg"
           style={{ padding: 20, background: 'rgba(255, 255, 255, 0.08)' }}
           multiple={false}
+          onRemove={() => {
+            setMainFile(undefined);
+            setCoverFile(undefined);
```
b2kdaman, line 403: Use `null`.

b2kdaman: Setting vairables as `undefined` is a bad pattern.

austbot (approved):

PR #1234, opened 2021-12-16
Fix for auction primary sales distribution

kevinrodriguez-io:
# Fix for auction primary sales distribution

The problem this solves is that pretty much on a multi-creator auction (let's use 3 as an example) where you have a 34, 33, 33 percent royalty distribution, the auction manager would close using the secondary sales percentage instead of the full 100%, giving the first seller a 93.4% of the funds and the rest a 3.4% or less approx. This occurs because the empty payment action looks at the metadata primary_sale_happened when deciding to do normal or secondary sale distribution. The primary sale flag is flipped when the user claims the item which often happens before funds are settled leading to secondary sale payout instead of primary.

## Steps to reproduce:

- 1. Create a multi creator NFT (34%, 33%, 33%).
- 2. Everyone signs it so you can list.
- 3. List it (Can be a 5m sale or an instant sale).
- 4. Buy with another account.
- 5. Get back to the original account and claim funds.
- 6. It will allocate funds in an unexpected way.

## Changes

Include an additional field on the safety deposit config snapshots the primary sale happened flag at the time of the listing creation. Then in empty payment evaluate the safe deposit config for is_primary_sale instead of the metadata since it was likely switched by the user redeeming the bid.

### DIFF
--- a/rust/metaplex/program/src/processor/empty_payment_account.rs
+++ b/rust/metaplex/program/src/processor/empty_payment_account.rs
@@ -6,7 +6,7 @@ use {
         instruction::EmptyPaymentAccountArgs,
         state::{
             get_auction_manager, AuctionManager, Key, PayoutTicket, Store, MAX_PAYOUT_TICKET_SIZE,
-            PREFIX, TOTALS,
+            PREFIX, TOTALS, SafetyDepositConfigV2, SafetyDepositConfig
         },
         utils::{
             assert_derivation, assert_initialized, assert_is_ata, assert_owned_by,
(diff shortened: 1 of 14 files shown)


rust/metaplex/program/src/processor/empty_payment_account.rs
```diff
@@ -85,11 +85,18 @@ fn calculate_owed_amount(
     winning_config_item_index: &Option<u8>,
     creator_index: &Option<u8>,
 ) -> Result<u64, ProgramError> {
-    let primary_sale_happened = auction_manager.get_primary_sale_happened(
-        metadata,
-        *winning_config_index,
-        *winning_config_item_index,
-    )?;
+    let primary_sale_happened = match safety_deposit_config_info.and_then(|a| {
```
kespinola, line 88: This is the primary change. It switches to reading primary sale happened from the metadata account to the snapshot of primary sale happened saved to the safety deposit config at the time of listing.

It is handling checking for v1 and v2 configs without introducing a new empty payment instruction.

kespinola (approved):

kespinola (commented):
If anyone has recommendations on the follow please let us know:

1. How to reduce the code duplication on rust/metaplex/program/src/processor/validate_safety_deposit_box_v3.rs and rust/metaplex/program/src/state.rs. Ideally like to a higher order struct that capable of processing configs v1 and v2.
2. Spec examples primarily around scaffolding auctions. The fixtures for this spec should be a metadata up for sale with multiple co-creators. Test split works when auction settled after the  nft claimed by the buyer.

PR #1268, opened 2021-12-20
Display countdown when no FLP ID is specified and CM has not started

candymandev:
There is a bug in the FLP frontend that shows nothing when no `REACT_APP_FAIR_LAUNCH_ID` is specified and it is before the candy machine launch date.

Current behavior:
bad

Behavior after this change:
good

If anyone has suggestions to fix the jarring display change, please let me know. I didn't want to hold up this fix for that issue, but it would be nice to fix that too!


roederw:
Fair Launch is being deprecated. Closing this.

PR #1270, opened 2021-12-20
Candymachine: Relax instruction check.

dankelleher:
CM uses the instruction sysvar to check for suspicious instructions in the CM transaction. For example, instructions that short-circuit the transaction if the minted NFT is not rare.

The present code prevents all transactions with unrecognised instructions. In fact, the CM should only be interested in instructions that come after the Mint instruction, as these are the only ones that have access to the minted nft account.

By relaxing this rule, transactions including CM mints can be composed with other instructions, that add features to CM, such as TokenGuard.

This PR allows any instructions in the tx before the CM instruction, while still restricting instructions after it.

### DIFF
--- a/rust/nft-candy-machine-v2/src/lib.rs
+++ b/rust/nft-candy-machine-v2/src/lib.rs
@@ -12,7 +12,10 @@ use {
             log::sol_log_compute_units,
             program::{invoke, invoke_signed},
             serialize_utils::{read_pubkey, read_u16},
-            system_instruction, sysvar,
+            system_instruction, sysvar::{
+                self,
+                instructions::load_current_index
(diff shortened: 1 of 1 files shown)


bhgames:
Not true. Any instruction BEFORE the CM can access the recent blockhash, which would allow them to figure out what WILL be minted.

dankelleher:
Damn. I was hoping that wouldn't be the case. So CM is just fundamentally uncomposable in its current form then?

bhgames:
Yep. Composability is a vulnerability for something like this. I could see a future CMv3 where you can whitelist certain programs for a specific candy machine, but that was not included in this version.

PR #1292, opened 2021-12-22
remove import of StringPublicKey that does not exist

garrettbear:
Build failing from StringPublicKey into borsh.ts and it's not needed.


garrettbear:
ahhhh booo. always a different error production 🥲

wagmi-cafe:
image
 had this issue too. fixed by removing these as well as the line you removed

PR #1302, opened 2021-12-24
[CLI] Create a config generator

vdusart:
Create python script to customize the upload config file


roederw:
I'm not sure we want this contained within the Metaplex repo. Could be worthwhile to just publish this under your own name and share with the world, should they choose to use it.

vdusart:
@roederw thank you for taking the time to look at my pr.
But I don't agree with you because in the repo you can find the file "psd_layer_generator.py" which allows to generate images. I don't really see how the generation of the configuration is different.

PR #1350, opened 2021-12-29
Separate dependencies fair-launch and token-entangler

febo:
The mint page for the Candy Machine v2 in the fair-launch package has conflicting dependencies with token-entangler package.


samuelvanderwaal (approved):

febo:
Merged on the release branch.

PR #1358, opened 2021-12-30
Fixed web dependencies 

febo:
* Added missing dependencies to gumdrop and web packages
* Filter fair-launch from build to avoid dependencies conflict


zaxozhu (approved):

febo:
Still having issues with libraries.

PR #1203, opened 2021-12-13
New candy machine

bhgames:

PR #1363, opened 2021-12-31
candy-machine-v2-cli: Fix `get_all_mint_addresses` cmd

APTy:
Currently, the `get_all_mint_addresses` command returns an empty
list. This has been noted by some users on discord already, but it
doesn't look like an issue has been filed yet.

This PR now properly derives the candy machine v2 address (as in
[metaboss/src/snapshot.rs](https://github.com/samuelvanderwaal/metaboss/blob/edeb9acdb63dc53278c66ffec4d0509b8304c5b7/src/snapshot.rs#L138)).

Usage:
```bash
$ ts-node ./candy-machine-v2-cli.ts get_all_mint_addresses \
    -e devnet -k ~/.config/solana/devnet.json -c cache
wallet public key: 26R5F...
[
  "5hA1Q...",
  "Edy2C..."
]
```

### DIFF
--- a/js/packages/cli/src/candy-machine-v2-cli.ts
+++ b/js/packages/cli/src/candy-machine-v2-cli.ts
@@ -24,6 +24,7 @@ import {
   loadCandyProgramV2,
   loadWalletKey,
   AccountAndPubkey,
+  deriveCandyMachineV2ProgramAddress,
 } from './helpers/accounts';
 
 import { uploadV2 } from './commands/upload';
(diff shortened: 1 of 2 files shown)


APTy:
Maybe @bhgames can take a look? I see this command was added in #1203 with the rest of the `v2`, and maybe just hasn't seen a lot of use yet. BTW I just started using CMv2 on devnet and it's been great so far!

bhgames (approved):

PR #1366, opened 2021-12-31
Fix for hijack attack of candy machine

bhgames:


SeanOverton:
how does the hijack attack work here out of curiosity??

SeanOverton:
@bhgames ^^

goriunov:
I might be wrong but i think this PR has disabled CM v2 rent reduction functionality... And I would argue that it is important to have, not everyone has the funds to launch with 2 years rent. If i want to launch within 1 week and have less funds on hands right now i would prefer to reduce rent and actually pay that small amount then try to get the fund to just have them locked up in CM.

For example for our collection to launch with 2 years rent we will need 15 SOL+ which is quite a big amount to pull in at once and will require some time to pull so our release will be pushed away ! 

At least you should provide some options to configure the rent and explain people which one means what and that if they go with 2 year they will pay nothing but need more funds and if they go with less than that they will actually lose some of the SOL locked up.

SeanOverton:
https://medium.com/@contact_94841/smashing-the-candy-machine-for-fun-and-profit-a3bcc58d6c30

SeanOverton:
For anyone interested

PR #1368, opened 2022-01-01
Fix bug with upload rerun

goriunov:
Basically current upoload will always try to reupload to Arweave even if all assets have been properly uploaded. In this case it will never try to move forward to configure CMv2 if for any reason it fails in the middle of the CMv2 configuration. This PR basically skips the Arweave  upload if there is no element to upload..

### DIFF
--- a/js/packages/cli/src/commands/upload.ts
+++ b/js/packages/cli/src/commands/upload.ts
@@ -107,99 +107,102 @@ export async function uploadV2({
     storage === StorageType.ArweaveBundle ||
     storage === StorageType.ArweaveSol
   ) {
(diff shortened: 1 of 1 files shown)


samuelvanderwaal:
@goriunov Thanks for the PR. Can you make it against `release/v1.1.0` instead of `master`. I want to merge the fix into our release branch that we're preparing.

goriunov:
@samuelvanderwaal I have created PR which points to the release/1.1.0 also have improved the fix to work for all storages and not just arweave-sol check https://github.com/metaplex-foundation/metaplex/pull/1382

PR #1386, opened 2022-01-04
Deprecated candy machine v1 cli

stegaBOB:


roederw:
I only had the nits above (and we are chatting offline in a couple hours). @samuelvanderwaal if you don't have any issues once we're done with the string _nits_ this should be good to go.

js/packages/cli/src/candy-machine-v1-cli.ts
```diff
@@ -362,12 +351,13 @@ programCommand('verify_token_metadata')
     log.info(`Elapsed time: ${timeTaken}\n`);
   });
 
-programCommand('verify')
+programCommand('verify_upload')
   .option(
     '-r, --rpc-url <string>',
     'custom rpc url since this is a heavy command',
   )
   .action(async (directory, cmd) => {
```
roederw, line 359: Are these being used?

samuelvanderwaal: Good question. I don't think this command can be used any more since the deprecation.

stegaBOB: The upload command can still be used to modify the config account. This would just verify that it modifies it correctly

stegaBOB: Perhaps altering the command name to something like `verify_config_account` or something like that would be more clear?

samuelvanderwaal:
> I only had the nits above (and we are chatting offline in a couple hours). @samuelvanderwaal if you don't have any issues once we're done with the string _nits_ this should be good to go.

Great, thanks. Don't let me be the blocker on this. I did a quick review and it all looked reasonable so once the minor issues are squared away you guys can merge this as far as I'm concerned.

js/packages/cli/src/candy-machine-v1-cli.ts
roederw: It's probably fine for now. I'm not going to dig too deep into V1 stuff. I assume we're just going to out right delete these sometime soon.

roederw (approved):
Ship it!

js/packages/cli/src/candy-machine-v1-cli.ts
```diff
@@ -983,12 +856,16 @@ programCommand('update_existing_nfts_from_latest_cache_file')
...
```
roederw, line 858: @stegaBOB

PR #1426, opened 2022-01-06
Skipping unsupported files instead of throwing an error

roederw:
Users are complaining that we cancel a Candy Machine upload when unexpected files are in their `assets` directory. Seems like a reasonable change to skip, inform and move on.


samuelvanderwaal (approved):

samuelvanderwaal:
The code looks good, not sure what's the issue with the Vercel builds.

PR #1431, opened 2022-01-06
Hidden Setting Limits

austbot:
Adds a check to not allow hidden settings to be toggled once you haveset a configline out of hidden settings.

Problem:
You could change items available even after adding items to create confusion and potentially malicious behavior.

Solution:

Limit hidden settings after a config line has been added without hidden settings on


aheckmann (commented):
code LGTM. how was this tested?

aheckmann:
FYI It is safe to ignore the candy machine ui CI failure as @roederw is iterating on it in another branch

roederw:
> FYI It is safe to ignore the candy machine ui CI failure as @roederw is iterating on it in another branch

I'm just going to try and fix it now (since this change might sit in 1.1.0 for a couple of days).

Issue #1322, opened 2021-12-26
Cannot read properties of undefined (reading 'slice')


calcajack3:
##  Cannot read properties of undefined (reading 'slice')

I was following the 'candy machine V2' tutorial on the metaplex docs to create a fair launch

Every step worked since the last one, when the guide says to run `yarn install` and `yarn start`
when the browser opens it gave me this error 
![image](https://user-images.githubusercontent.com/96479697/147417539-eaeb8078-56f8-4956-933a-346b7617a96e.png)

I have done all the previous step exactly

**Desktop**
 - OS: Windows 11
 - Browser: Brave
- Version: 1.33.106 (Chromium: 96)

limjame28:
i have the same issue here

ludusdigital:
Try to run **npm install**

devansh-chotai:
> Try to run **npm install**

Thanks! that works

calcajack3:
> Try to run **npm install**

i did it but it tells me that is already installed

Blueberry94:
For starting the server in development mode you should go with npm start. 

Worked for me too. Thanks a lot!

HaoCherHong:
This is due to fair launch is depending on `@solana/wallet-adapter-react@0.13.1`
while some other dependencies are using `@solana/wallet-adapter-react@0.7.2`.
you can see that by using the following command:
```
yarn list @solana/wallet-adapter-react
```
multiple instances of `wallet-adapter-react` is created, so `WalletProvider` is not working.

ck256-2000:
> Try to run **npm install**

thanks for this  - worked on my end as well!

PR #1440, opened 2022-01-07
Fix fair-launch dependencies problem for `yarn`

HaoCherHong:
Since the docs recommends `yarn` to be used as package manager for this project,
In order to fix problem described in #1322 and [this comment](https://github.com/metaplex-foundation/metaplex/issues/1322#issuecomment-1006153851), we need to make sure yarn resolve all `@solana/wallet-adapter-react` to same version.
Unfortunately, yarn doesn't see "^0.7.2" and "^0.13.1" as compatible like npm does, I manually updated them all.

Another way to solve this is to enforce all `@solana/wallet-adapter-react` to be resolved to same version with [resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-how-to-use-it) in `package.json` file. But I think it's a workaround and not a long term solution.

# changes
Upgrade all `@solana/wallet-adapter-react` to `^0.13.1`

### DIFF
--- a/js/packages/common/package.json
+++ b/js/packages/common/package.json
@@ -36,7 +36,7 @@
     "@solana/spl-token": "0.1.6",
     "@solana/spl-token-registry": "0.2.202",
     "@solana/wallet-adapter-base": "^0.4.1",
-    "@solana/wallet-adapter-react": "^0.7.2",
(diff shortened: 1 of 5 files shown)


roederw:
This is already fixed in `release/v1.1.0` and will get pulled into `master` soon. Closing.

PR #1444, opened 2022-01-07
bump @civic/solana-gateway-react to  0.4.7

kevinhcolgan:
Bump civic react component to 0.4.7 that uses new failover-tolerant Civic Gatekeeper endpoints


febo (changes requested):
The @civic/solana-gateway-react version was updated in the new candy-machine-ui package, which replaced the fair-launch package. The only change needed from this PR is the update to the readme file.

Issue #1306, opened 2021-12-25
Fair Launch TypeError. Doesnt launch correctly.


Fiercevd:
Fair Launch isn't work correctly.

When yarn start in the FLP folder there is an issue:

TypeError: Cannot read properties of undefined (reading 'slice')

![image](https://user-images.githubusercontent.com/96639493/147375618-1db97cb3-171a-427d-97f9-f6dfaffdf057.png)
![image](https://user-images.githubusercontent.com/96639493/147375625-92d5f8a2-7df6-4c8a-a306-6ddc02b9b0a3.png)

catsonmats:
Also receiving this error. My steps:

1. Successfully run the upload command from the v2 cli
2. Input candy machine ID into the .env
3. Uncomment devnet options in .env, comment out mainnet options
4. yarn start
5. Successful start of dev server, try to reach localhost:3000, receive error

guillemavila:
The problem comes when loading the canvas-confetti module. 
Don't want to submit a PR as this is not a _real_ fix, but you can circumvent the compiling error by editing the file `~/[...]/fair-launch/src/confetti.tsx` and adding `// @ts-ignore` right before `import confetti from 'canvas-confetti';`:

```
import confetti from 'canvas-confetti';    ->    // @ts-ignore
                                                 import confetti from 'canvas-confetti';
```

trimpage:
@guillemavila i tried this but it still doesn't work

Aaron-ATL:
@guillemavila @trimpage I am having the exact same issue. `Cannot read properties of undefined (reading 'slice')`

@guillemavila's fix did not work unfortunately. Why is this happening?

JFK344:
Anyone managed to get it started ?

akserikawa:
Having the same problem. @guillemavila fix doesn't work. Do you have any other ideas about what it might be causing this?

limjame28:
im having the same trouble here

epomatti:
same same, fresh from `master`

![image](https://user-images.githubusercontent.com/6530769/147439026-48e0814a-0350-4c45-b17f-95a44567bada.png)

marcelc63:
Error comes from other folders, not fair-launch folder. Since its a monorepo ts checks for everything it seems.

Solution:
Copy paste fair-launch to a different directory and build separately

or checkout to this commit 56de262a3ea3f0f47d2f98d270e95085b25739b8

That works for me.

epomatti:
[...]

PR #1453, opened 2022-01-07
Update Fair-Launch dev canvas-confetti dependency

MonarchButterflySquad:
Added a missing dev dependency @types/canvas-confetti, should solve https://github.com/metaplex-foundation/metaplex/issues/1306#issuecomment-1002654316.


roederw:
`fair-launch` has been removed.

PR #1458, opened 2022-01-08
Bug fix  "Can not parse presale of null" CM UI

goriunov:
Bug fix of the @febo presale implementation https://github.com/metaplex-foundation/metaplex/pull/1456, where it will not allow to mint for non presale...


AlmostEfficient:
This worked for me, thanks a bunch!

stegaBOB (approved):

PR #1470, opened 2022-01-08
 bugfix to PR #1456

halaprix:
existence of `state.data.whitelistMintSettings` has not been checked - that caused an error while getting candy machine state.


what-name:
Screenshot 2022-01-08 at 23 28 58

goriunov:
There is already open pr here https://github.com/metaplex-foundation/metaplex/pull/1458

PR #1497, opened 2022-01-10
[Candy Machine UI] Presale price correction

roederw:
- The presale price wasn't being displayed, instead only the live mint price was. Corrected this problem.
- Added UI queue that you're minting in a environment "presale"

image


stegaBOB (approved):

epomatti:
This is looking good 🍡

PR #1500, opened 2022-01-10
Make bundlr upload (arweave-sol) retry instead of ungraceful fail

goriunov:
At the moment if for any reasons arweave-sol upload fails including the network problems/timeouts and the actual fund function call succeeds people will lose their money, this PR just adds simple retry mechanism to the bundlr upload to reduce the chance of losing funds due to occasional instability issues.

The tx upload will be attempted 3 times with small sleep of 10s between each attempt if none of the 3 times succeed it will ungraceful fail (fall back to old behavior)

Ideally we should be looking in to adding arweave fund tx id to the cache and reuse it on new reuploads if possible to completely or in most cases reduce the chance of losing funds (something to think about for later)

### DIFF
--- a/js/packages/cli/src/commands/upload.ts
+++ b/js/packages/cli/src/commands/upload.ts
@@ -231,14 +231,6 @@ export async function uploadV2({
                 allIndexesInSlice[i] >= lastPrinted + tick ||
                 allIndexesInSlice[i] === 0
               ) {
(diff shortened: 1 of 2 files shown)


roederw:
Thanks for the improvement. Looks good to me!

roederw (approved):

PR #926, opened 2021-11-08
feat: add nft.storage upload option

yusefnapora:
Hey there, Yusef from Protocol Labs here 😄. I've been working on getting Metaplex and [NFT.Storage](https://nft.storage) to play nicely together for a bit, and it seems close enough to get some feedback on. Some things are not yet wired up in our backend, so I'm just opening as a draft for now. 

I'm planning to also add this functionality to the new [JS SDK](https://github.com/metaplex/js), but it seemed easier to test things out here.

The main work of the integration is happening in the [metaplex-dotstorage-auth](https://github.com/nftstorage/metaplex-dotstorage-auth) package that I've been writing up. It's not yet published to NPM, so there's currently a github dependency in the package.json for the candy machine cli package. That will be removed before this is converted from a draft PR.

The auth works entirely locally - the user just creates a self-signed JWT using the `EdDSA` algorithm with their private Solana Ed25519 key as the signing key. There's no Solana transaction required, just the signature and a small payload embedded in the JWT that looks like this:

```json
{
  "iss": "did:key:<users-public-key>",
  "req": {
    "put": {
      "rootCID": "bafy123...",
      "tags": {
        "chain": "solana",
        "solana-cluster": "devnet"
      }
    }
  }
}
```

The JWT payload includes the Metaplex user's public key (encoded as a DID) in the `iss` field.  The `req` field contains a description of the request, including the root CID of the content being uploaded and some tags to help keep track of things on our end.

The CID is calculated on the client by packing files into CAR archives, which are sent to the nft.storage backend in chunks of ~10MB at a time.

### Status

There are some remaining pieces needed before this will be functional:

- [-] nft.storage backend updated to allow uploads with `X-Web3-Auth: Metaplex <token>` header support
  - currently deployed to staging environment. See below for usage
- [x] metaplex-dotstorage-auth library published to NPM

I'll convert the PR from draft once it's possible to actually test this end to end. In the meantime, if any maintainers here have input on the code or any concerns, let me know and I'll try to address them ASAP.

### Update (2021-11-23):

We've now published [`@nftstorage/metaplex-auth`](https://github.com/nftstorage/metaplex-auth) to npm, so I've updated this PR to depend on it.

The API changes are currently deployed to our staging environment, but not yet in production. To test, you can set the `NFT_STORAGE_ENDPOINT` environment variable to `https://api-staging.nft.storage` before running the `candy-machine upload` command.

For example:

```
[...]

PR #1512, opened 2022-01-11
Feature/add nft storage

kjaylee:
Add storage.  https://nft.storage


yusefnapora:
Yay! Glad to see nft.storage and candy machine together at last :)

Just wanted to flag that using the HTTP api directly in this way has a 100MiB file size limit. You can avoid that by using the nft.storage client package, which encodes files into CAR archives and chunks them before uploading.

Also, let me know if there's any interest in adding the wallet-based auth mechanism used by #926, which doesn't require the user to have an NFT.Storage account.

PR #1515, opened 2022-01-11
fixed the keypair requirement for offline commands

halaprix:
keypair has been required for all commands including offline ones like "create_generative_art". It has been introduced in PR #1409


stegaBOB:
I think there is only one command that doesn't require a keypair as of now. For error handling, I think its better to keep it as a required option for the benefit of all the other commands.

stegaBOB:
Actually it looks like there are a bit more than 1. Going to make the required option on the needed commands

PR #1527, opened 2022-01-12
fixed `update_existing_nfts_from_latest_cache_file` in cm v2

halaprix:
the `update_existing_nfts_from_latest_cache_file` command in v2 has not been wokring because it assumed the first creator of the mint is the cm account. Based on that the metadata lookup has been performed. I've changed it to the address derived from `candy_machine` and cm address seed.

### DIFF
--- a/js/packages/cli/src/commands/updateFromCache.ts
+++ b/js/packages/cli/src/commands/updateFromCache.ts
@@ -62,6 +62,15 @@ export async function updateMetadataFromCache(
   cacheContent: any,
   newCacheContent: any,
 ) {
+  const utf8Encode = new TextEncoder();
+  const temp = await PublicKey.findProgramAddress(
+    [
+      utf8Encode.encode('candy_machine'),
+      new PublicKey(candyMachineAddress).toBytes(),
+    ],
+    new PublicKey('cndy3Z4yapfJBmL3ShUp5exZKqR3z33thTzeNMm2gRZ'),
+  );
+  candyMachineAddress = temp[0].toString();
   const metadataByCandyMachine = await getAccountsByCreatorAddress(
     candyMachineAddress,
     connection,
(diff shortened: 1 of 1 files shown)


MyNameisLeon:
I tested the command and it is working fine for me with this changes. Would be great to have it on master! @samuelvanderwaal

js/packages/cli/src/commands/updateFromCache.ts
```diff
@@ -62,6 +62,15 @@ export async function updateMetadataFromCache(
   cacheContent: any,
   newCacheContent: any,
 ) {
+  const utf8Encode = new TextEncoder();
+  const temp = await PublicKey.findProgramAddress(
+    [
+      utf8Encode.encode('candy_machine'),
+      new PublicKey(candyMachineAddress).toBytes(),
+    ],
+    new PublicKey('cndy3Z4yapfJBmL3ShUp5exZKqR3z33thTzeNMm2gRZ'),
```
stranzhay, line 71: if you could update 71 and 68 to use the constants for these values, this will ensure smoother changes in the future for any updates.

js/packages/cli/src/commands/updateFromCache.ts
```diff
@@ -62,6 +62,15 @@ export async function updateMetadataFromCache(
   cacheContent: any,
   newCacheContent: any,
 ) {
+  const utf8Encode = new TextEncoder();
```
stranzhay, line 65: i think even more so could `deriveCandyMachineV2ProgramAddress` be used in this case @samuelvanderwaal

js/packages/cli/src/commands/updateFromCache.ts
austbot: agreed there are constants that you can invoice

halaprix:
updated with deriveCandyMachineV2ProgramAddress

stranzhay (approved):

PR #1528, opened 2022-01-12
fixed presale price when no presale discount

halaprix:
It has been assumend that when presale is active, there is a discounted price.
Check if the discounted price exists has been added


roederw (approved):

roederw:
Thanks for the fix! Ship it!

PR #1540, opened 2022-01-13
Fix serious verify command bug on rerun with partially verified cache

goriunov:
At the moment verify command is broken... it only works if all the assets verified at the same time. If you only were able to verify first set of assets in your cache and not the second one when you try to rerun verify command it will fail as the index will be pointing from 0 for your second set and the addresses and names it is pointing will not math the reality.

This might bring alot of confusions as after you run verify command and get some parts of your collection unverified and then trying uploading again and run second time verify it will fail and reset your just uploaded indices back to onChain false

This PR corrects the verify command to actually look through affected keys instead of indexes of the entries in the array. Have also added small improvement to reset verifyRun to false if we write indices for that specific entry.

also enable logging to better express the command execution

### DIFF
--- a/js/packages/cli/src/candy-machine-v2-cli.ts
+++ b/js/packages/cli/src/candy-machine-v2-cli.ts
@@ -358,46 +358,42 @@ programCommand('verify_upload')
     );
     let allGood = true;
 
-    const keys = Object.keys(cacheContent.items).filter(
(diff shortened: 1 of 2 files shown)


roederw (approved):

roederw:
Good to go! Thanks for the clean up as well!

PR #1552, opened 2022-01-14
Added edition number option to minting art

yamijuan:
image


roederw (approved):

roederw:
What happens if `editions > 1`. In that scenario, what happens to `Edition Number`?

yamijuan:
Hey so this is kinda a hotfix for some cases where the edition number does not get minted but the counter somehow got updated, so people can fix missing editions

YoungMagicalMan (approved):

PR #1554, opened 2022-01-14
fixed `Class UpdateMetadataV2Args is missing in schema`

halaprix:
added `UpdateMetadataV2Args` and `DataV2` classes that were missing in `schema.ts` that preventted running the update metadata command in `cli-nft.ts`


roederw:
Actually, I think I'm going to remove this schema file and just replace it with `token-metadata`

21e8:
isn't that closed-source?

PR #1556, opened 2022-01-14
Allow CM creation from custom cache file with custom links and skipping upload process

goriunov:
This PR will allow creation of the CM from the cache file which can point to any website which is hosting their own metadata. Basic steps is to create custom cache with links and onChain: false in it which point to off chain metadata.

The steps will be CM will take 0.json file from assets to pull some metadata for onchain metadata for CM then the upload will be skipped as all the links already have been added to the cache file. The next step script will be doing is writing indices to the candy machine and done you have CM ready !

Custom cache should have program set to empty object before the first upload.

### DIFF
--- a/js/packages/cli/src/commands/upload.ts
+++ b/js/packages/cli/src/commands/upload.ts
@@ -100,13 +100,7 @@ export async function uploadV2({
     : undefined;
 
   if (!cacheContent.program.uuid) {
-    const firstAssetKey = dedupedAssetKeys[0];
-    const firstAssetManifest = getAssetManifest(
-      dirname,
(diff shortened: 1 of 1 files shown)


goriunov:
@samuelvanderwaal @stegaBOB wondering if we can merge this PR in... One of the examples where it would work well..

if you use arloader to upload all your nfts and then create custom cache for the CM it wont be able to create CM without this PR because when it creates CM it needs at least one item not uploaded as it is looking for the not uploaded items... 

This PR will always look at the item 0 and should not bother if the item was uploaded or not

samuelvanderwaal (approved):

PR #1569, opened 2022-01-16
Update ARCHITECTURE.md

moflo:
Adding correct link to instruction set for the token metadata contract

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -178,7 +178,7 @@ pub struct EditionMarker {
 
 ```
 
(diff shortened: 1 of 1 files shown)


samuelvanderwaal:
Good catch, but any reason you pointed to the state instead of the instruction in the new repo?

moflo:
The state definition seemed more relevant to my needs as I was seeking information on the protocol, but I understand you previously had the instruction linked.

samuelvanderwaal (approved):

PR #1572, opened 2022-01-17
Candy Machine UI: update UI to COMPLETED when end date reached

cosimo-rip:
Follow on to #1571 - when we hit the end date set for a candy machine, the UI doesn't take any action.

This PR sets a timer to refresh the candy machine if the browser window is still open when the end date is reached. This will cause the isActive state to be updated and the UI to show the COMPLETED state.

Note that this is a best effort attempt. The frontend UI and the blockchain clocks _could_ be out of sync to some degree, so for that reason I introduced a 5s delay before attempting the update.

### DIFF
--- a/js/packages/candy-machine-ui/src/Home.tsx
+++ b/js/packages/candy-machine-ui/src/Home.tsx
@@ -161,6 +161,25 @@ const Home = (props: HomeProps) => {
     refreshCandyMachineState,
   ]);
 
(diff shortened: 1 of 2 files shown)


roederw:
We have plans refresh all UI information based on a Candy Machine state update, using only one hook. Closing this since it'l become redundant once the change goes in.

PR #1574, opened 2022-01-17
fix: video not showing in the homepage cards

yamijuan:
# Problem
If the primary NFT is a video, they won't show in the homepage cards

# Solution
This is a css Issue so it was solved by displaying the video container correctly. This was tested by creating an auction/sale with a video NFT and make sure the video is played in the homepage auction card

Before:
image

After: 
video


aheckmann:
Please add screen shots demonstrating the improvement

yamijuan:
Done @aheckmann

aheckmann (approved):

PR #1607, opened 2022-01-20
feat: add max supply and verify creators options to nft cli

jon-nona:
# Pull Request - Metaplex

## Scope

- Add max supply and verify creators options to js/packages/cli/src/cli-nft.ts mint command
- Add verify creators option to js/packages/cli/src/cli-nft.ts update-metadata command

## Work Done

- Add max supply and verify creators options to js/packages/cli/src/cli-nft.ts mint command
- Add verify creators option to js/packages/cli/src/cli-nft.ts update-metadata command

## Steps to Test

- from js/packages/cli/src run (assuming that you have your keypair defined and metadata set up):

```
ts-node cli-nft.ts mint --env devnet --keypair <PATH-TO-KEYPAIR-JSON>  --url <URL-TO-METADATA> --max-supply 5 --verify-creators 0
```

and for update metadata:

```
ts-node cli-nft.ts update-metadata --env devnet --keypair <PATH-TO-KEYPAIR-JSON>  --mint <BASE-58-MINT-KEY> --url <URL-TO-METADATA> --verify-creators 0
```

### DIFF
--- a/js/packages/cli/src/cli-nft.ts
+++ b/js/packages/cli/src/cli-nft.ts
@@ -14,13 +14,24 @@ programCommand('mint')
   .option('-u, --url <string>', 'metadata url')
   .option(
     '-c, --collection <string>',
-    'Optional: Set this NFT as a part of a collection, Note you must be updat authority for this to work.',
+    'Optional: Set this NFT as a part of a collection, Note you must be the update authority for this to work.',
   )
   .option('-um, --use-method <string>', 'Optional: Single, Multiple, or Burn')
   .option('-tum, --total-uses <number>', 'Optional: Allowed Number of Uses')
+  .option('-ms, --max-supply <number>', 'Optional: Max Supply')
+  .option('-vc, --verify-creators <number>', 'Optional: 0 or 1')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   .action(async (directory, cmd) => {
-    const { keypair, env, url, collection, useMethod, totalUses } = cmd.opts();
+    const {
+      keypair,
+      env,
+      url,
+      collection,
+      useMethod,
+      totalUses,
+      maxSupply,
+      verifyCreators,
+    } = cmd.opts();
     const solConnection = new web3.Connection(getCluster(env));
     let structuredUseMethod;
     try {
@@ -33,12 +44,16 @@ programCommand('mint')
     if (collection !== undefined) {
       collectionKey = new PublicKey(collection);
     }
+    const supply = maxSupply || 0;
+    const verify = typeof verifyCreators === 'undefined' ? 1 : verifyCreators;
     await mintNFT(
       solConnection,
       walletKeyPair,
       url,
       true,
       collectionKey,
+      supply,
+      verify,
       structuredUseMethod,
     );
   });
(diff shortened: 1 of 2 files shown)


js/packages/cli/src/cli-nft.ts
```diff
@@ -52,10 +67,19 @@ programCommand('update-metadata')
   )
   .option('-um, --use-method <string>', 'Optional: Single, Multiple, or Burn')
   .option('-tum, --total-uses <number>', 'Optional: Allowed Number of Uses')
+  .option('-vc, --verify-creators <number>', 'Optional: 0 or 1')
```
austbot, line 70: I think this would be better as a bool do you agree?

js/packages/cli/src/cli-nft.ts
```diff
@@ -14,13 +14,24 @@ programCommand('mint')
   .option('-u, --url <string>', 'metadata url')
   .option(
     '-c, --collection <string>',
-    'Optional: Set this NFT as a part of a collection, Note you must be updat authority for this to work.',
+    'Optional: Set this NFT as a part of a collection, Note you must be the update authority for this to work.',
   )
   .option('-um, --use-method <string>', 'Optional: Single, Multiple, or Burn')
   .option('-tum, --total-uses <number>', 'Optional: Allowed Number of Uses')
+  .option('-ms, --max-supply <number>', 'Optional: Max Supply')
```
austbot, line 21: very nice

js/packages/cli/src/cli-nft.ts
```diff
@@ -14,13 +14,24 @@ programCommand('mint')
   .option('-u, --url <string>', 'metadata url')
   .option(
     '-c, --collection <string>',
-    'Optional: Set this NFT as a part of a collection, Note you must be updat authority for this to work.',
+    'Optional: Set this NFT as a part of a collection, Note you must be the update authority for this to work.',
```
austbot, line 17: thanks

austbot (changes requested):

js/packages/cli/src/cli-nft.ts
jon-nona: @austbot yes I do for sure. Will make that change in a little.

js/packages/cli/src/cli-nft.ts
jon-nona: Thank you :)

jon-nona:
@austbot I think I just missed the bool option in types in the cli. I will amend and push and re-request as soon as I have.

jon-nona:
@austbot from [this](https://www.npmjs.com/package/commander#common-option-types-boolean-and-value)

>Other option types, negatable boolean and boolean|value
>You can define a boolean option long name with a leading no- to set the option value to false when used. Defined alone this also makes the option true by default.

So maybe have: 

```
.option('-nvc, --no-verify-creators, 'Optional: disable verification of creators')
```

verify creators as I read it would thus be true unless you pass it in the cli. If you like that approach then is nvc for the short flag alright?

irs-team (approved):
Thank you for this! Works fantastic 🙏

austbot (approved):

js/packages/cli/src/commands/mint-nft.ts
```diff
@@ -75,7 +76,7 @@ export const createMetadata = async (
       new Creator({
         address: creator.address,
         share: creator.share,
-        verified: 1,
+        verified: verifyCreators ? 1 : 0,
```
oboxodo, line 79: @austbot assuming `verifyCreators` is `true` and there are creators other than the one running the command, this will fail trying to verify ALL of them at once, right? The program won't allow the creator running the command to sign/verify the other creators.

Shouldn't the `verifyCreators` (and `--no-verify-creators` option) be changed to singular and then this line be changed to something like the following?

(untested)
```suggestion
        verified: (creator.address = wallet.publicKey && verifyCreators) ? 1 : 0,
```

PR #1617, opened 2022-01-21
check if goLiveDate is not `null`

halaprix:
getCandyMachineState failed when `goLiveDate ` was not set.
Now it will set `isActive` to `false`


goriunov (approved):

roederw:
Can you provide more logs to describe exactly what you're seeing?

roederw:
Thanks for the fix!

roederw (approved):

PR #1620, opened 2022-01-21
fix(bidding): Revoke before canceling for placing bid

yamijuan:
# Problem
Due to a change in the core `spl_token` program. The order of clean-up instructions matters. This fix was rolled out and tested on the same codebase on holaplex.

# Solution 
Reverse the clean up instructions so the delegate is removed before the close account


aheckmann:
Please add a description and link to the open issue this PR fixes. Also, how did you test?

austbot (approved):

stetommy (commented):
Works

PR #1658, opened 2022-01-26
Respect chosen cluster when fetching tokens

oboxodo:
I'm not sure why the list of SPL Tokens was hardcoded to be fetched from `mainnet-beta`, instead of respecting the same endpoint being used in the global connection and chosen from the UI/LocalStorage/param. I asked about this in the #storefront Discord channel but didn't get a response.

With this PR, the list of SPL Tokens is filtered by whatever cluster is chosen by the user and not always `mainnet-beta`.

If there's a reason this was hardcoded maybe it's a good idea to add a comment in the code about it. But I saw no reason for it.

Most of the changes are just a rename of variables from `mainnetTokens` to `subscribedTokens` to make it less cluster-specific and more semantic.

The real change is to reuse the already filtered by chain id list of tokens coming from the connection instead of using a hardcoded filter by cluster name.

### DIFF
--- a/js/packages/web/src/contexts/tokenList.tsx
+++ b/js/packages/web/src/contexts/tokenList.tsx
@@ -13,7 +13,7 @@ export const SPL_REGISTRY_SOLLET_TAG = "wrapped-sollet";
 export const SPL_REGISTRY_WORM_TAG = "wormhole";
 
 export interface TokenListContextState {
-  mainnetTokens: TokenInfo[];
+  subscribedTokens: TokenInfo[];
   tokenMap: Map<string, TokenInfo>;
   wormholeMap: Map<string, TokenInfo>;
   solletMap: Map<string, TokenInfo>;
@@ -43,12 +43,12 @@ export function SPLTokenListProvider({ children = null as any }) {
   const hasOtherTokens = !!process.env.NEXT_SPL_TOKEN_MINTS;
 
   // Added tokenList to know in which currency the auction is (SOL or other SPL)
-  const mainnetTokens = tokenList?tokenList.filterByClusterSlug("mainnet-beta").getList().filter(f=> subscribedTokenMints.some(s=> s == f.address) )
+  const subscribedTokens = tokenList?tokenList.filterByClusterSlug("mainnet-beta").getList().filter(f=> subscribedTokenMints.some(s=> s == f.address) )
     :[]
 
   const tokenMap = useMemo(() => {
     const tokenMap = new Map();
-    mainnetTokens.forEach((t: TokenInfo) => {
+    subscribedTokens.forEach((t: TokenInfo) => {
       tokenMap.set(t.address, t);
     });
     return tokenMap;
@@ -56,7 +56,7 @@ export function SPLTokenListProvider({ children = null as any }) {
 
   // Tokens with USD(x) quoted markets.
   const swappableTokens = useMemo(() => {
-    const tokens = mainnetTokens.filter((t: TokenInfo) => {
+    const tokens = subscribedTokens.filter((t: TokenInfo) => {
       const isUsdxQuoted =
         t.extensions?.serumV3Usdt || t.extensions?.serumV3Usdc;
       return isUsdxQuoted;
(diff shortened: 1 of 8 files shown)


js/packages/web/src/contexts/tokenList.tsx
```diff
@@ -36,27 +36,29 @@ export function SPLTokenListProvider({ children = null as any }) {
       ...process.env.NEXT_SPL_TOKEN_MINTS.split(",")
     ]: [WRAPPED_SOL_MINT]
 
+  const { tokens } = useConnectionConfig();
+
   useEffect(() => {
-    new TokenListProvider().resolve().then(setTokenList);
-  }, [setTokenList]);
+    setTokenList(new TokenListContainer(Array.from(tokens.values())))
```
oboxodo, line 42: This is the main change to fetch the list of tokens from the already-filtered by chain id `tokens` (coming from the connection) instead of the full list of tokens.

js/packages/web/src/contexts/tokenList.tsx
```diff
@@ -36,27 +36,29 @@ export function SPLTokenListProvider({ children = null as any }) {
...
+  }, [setTokenList, tokens]);
 
   const hasOtherTokens = !!process.env.NEXT_SPL_TOKEN_MINTS;
 
   // Added tokenList to know in which currency the auction is (SOL or other SPL)
-  const mainnetTokens = tokenList?tokenList.filterByClusterSlug("mainnet-beta").getList().filter(f=> subscribedTokenMints.some(s=> s == f.address) )
+  const subscribedTokens = tokenList?tokenList.getList().filter(f=> subscribedTokenMints.some(s=> s == f.address) )
```
oboxodo, line 48: Removing the hard-coded `"mainnet-beta"` from here is the whole purpose of this PR. Now that the `tokenList` is already filtered using whatever cluster is chosen by the user, this can be removed.

stegaBOB:
Thanks for the PR! I'll review/test it more tomorrow, but it looks good in the quick glance I gave. Hopefully we'll get this merged by the end of the weekend.

stegaBOB:
@oboxodo looks like some recent PRs introduced some conflicts. Would you mind fixing those up for me? Once you do I'll merge it in.

oboxodo:
Thanks for the feedback, @stegaBOB. I've pushed a new PR in https://github.com/metaplex-foundation/metaplex/pull/1731.

I know this is only used to fetch a list of tokens as a "suggestion" for the user to use. But I still think it's confusing to have it hardcoded to mainnet. And if the suggested tokens have the same address on both clusters, then that's even more reason to merge this as it would work on both clusters anyway.

Honestly, my end goal is to allow devs implementing their own marketplace (like me) to be able to test on devnet with ANY custom token. We have our own custom token we'd like to use for selling NFTs and I'm having to paste the address in the input bellow the list instead of being able to pick from the list as I think I should. Even worse, my token's logo is not shown even when entering the address.

Metaplex's contracts are not limited to specific SPL Tokens. Why limit the UI?

Here's a small screencast of a custom token being shown on the list, after adding it to the `SPL_TOKEN_MINTS` env var in the `.env`:
picking_an_spl_token_from_devnet

DISCLAIMER: There's an extra change I had to do to make that work, which is not part of my PR. For some reason MEtaplex's UI limits the list only to tokens which have a market on USDT/USDC. Again... I think that's ok as a suggestion but why not allow store owners to use their own tokens even without that? But this is subject of a separate PR:

```diff
diff --git a/js/packages/web/src/contexts/tokenList.tsx b/js/packages/web/src/contexts/tokenList.tsx
index 7c2897a3..9166672d 100644
--- a/js/packages/web/src/contexts/tokenList.tsx
+++ b/js/packages/web/src/contexts/tokenList.tsx
@@ -58,11 +58,7 @@ export function SPLTokenListProvider({ children = null as any }) {
 
   // Tokens with USD(x) quoted markets.
   const swappableTokens = useMemo(() => {
-    const tokens = subscribedTokens.filter((t: TokenInfo) => {
-      const isUsdxQuoted =
-        t.extensions?.serumV3Usdt || t.extensions?.serumV3Usdc;
-      return isUsdxQuoted;
-    });
+    const tokens = subscribedTokens
     tokens.sort((a: TokenInfo, b: TokenInfo) =>
       a.symbol < b.symbol ? -1 : a.symbol > b.symbol ? 1 : 0
     );
```

Issue #1616, opened 2022-01-21
Error message RangeError: indeterminate span when uploading is not appropriate


saito-kosuke:
**Is your feature request related to a problem? Please describe.**

Error message `RangeError: indeterminate span` when uploading is not appropriate.
When there is no symbol in the metadata file, the error above has occurred.
In discord, many users of metaplex ask what this error message means.

**Describe the solution you'd like**

We should add the validation to the function createCandyMachineV2 like below.

```
  if (!candyData.symbol) {
    throw new Error(`Invalid config, there must be a symbol.`);
  }
```

I can make a pull request.
If you like, give me the write rights of this repository.

**Describe alternatives you've considered**

**Additional context**

Neoziion:
I got the same error message but I don't understand how to solve it. Can anyone help me?

saito-kosuke:
@Neoziion 
Maybe, the property `symbol` is missing from your JSON files.

RidhwanDev:
I also have this issue, but I do have symbol and I can see symbol if I do `console.log(candyData.symbol)` so I don't think it's exclusive to that property, maybe when other things are missing. I can't see what is required and what is optional though

saito-kosuke:
@RidhwanDev 
Let me see your JSON file to reproduce your issue.

> I can't see what is required and what is optional though

Totally agree.
We are providing a free tool called Coinfra to check the metadata. 
https://app.gitbook.com/s/t3DGhcnB6Csc5Eht4Jm7/tutorials/step2-option-how-to-check-assets-of-metaplex-candy-machine

stegaBOB:
@saito-kosuke Feel free to open a PR with your proposed fix! We'll merge that in (if it hasn't been added already). Thanks

saito-kosuke:
@stegaBOB 
Could you review my PR https://github.com/metaplex-foundation/metaplex/pull/1693 ?

DSellees:
What is the problem?

Beginning the upload for 1606 (img+json) pairs
started at: 1656495728060
initializing candy machine
Error deploying config to Solana network. RangeError: indeterminate span
    at Structure.getSpan

PR #1693, opened 2022-02-02
Add symbol validation to createCandyMachineV2

saito-kosuke:
fix: #1616

### DIFF
--- a/js/packages/cli/src/helpers/accounts.ts
+++ b/js/packages/cli/src/helpers/accounts.ts
@@ -121,6 +121,10 @@ export const createCandyMachineV2 = async function (
 ) {
   const candyAccount = Keypair.generate();
   candyData.uuid = uuidFromConfigPubkey(candyAccount.publicKey);
+  
+  if (!candyData.symbol) {
+    throw new Error(`Invalid config, there must be a symbol.`);
+  }
 
   if (!candyData.creators || candyData.creators.length === 0) {
     throw new Error(`Invalid config, there must be at least one creator.`);


stegaBOB (approved):
Looks good! Thanks

RidhwanDev:
I have symbol, this still fails.

saito-kosuke:
Could you share your config file?

RidhwanDev:
```{
    "name": "Number #0001",
    "symbol": "NB",
    "description": "Collection of 10 numbers on the blockchain. This is the number 1/10.",
    "seller_fee_basis_points": 500,
    "image": "0.png",
    "attributes": [
        {"trait_type": "Layer-1", "value": "0"},
        {"trait_type": "Layer-2", "value": "0"}, 
        {"trait_type": "Layer-3", "value": "0"},
        {"trait_type": "Layer-4", "value": "1"}
    ],
    "properties": {
        "creators": [{"address": "MY-ADDRESS", "share": 100}],
        "files": [{"uri": "0.png", "type": "image/png"}]
    },
    "collection": {"name": "numbers", "family": "numbers"}
}```

saito-kosuke:
@RidhwanDev looks good.
Could you check your assets via [this tool](https://support.coinfra.io/tutorials/step2-option-how-to-check-assets-of-metaplex-candy-machine) for free?

RidhwanDev:
@saito-kosuke I checked it on your tool, looks fine uploaded fine. Nice tool btw! I'm also trying to make candy-machine-v2 work from a react app. Can't get past this issue.

oboxodo:
@stegaBOB I think this commit should be reverted because it fails uploads when metadata has `symbol: ""` (empty string) but that value is actually valid. I could be wrong but there's nothing in the standard indicating a non-empty value MUST be used, and there's plenty valid NFTs in the wild with empty string symbols.

oboxodo:
@stegaBOB that shouldn't be necessary because Typescript already checks for a `string`.

stegaBOB:
Ah I think I misunderstood the issue then. It looks like we should probably change that type to be `string | undefined` and then convert that to `""` if undefined to satisfy the types needed for the candy machine on chain data.

oboxodo:
That works for me as long as I'm able to either avoid having a `symbol` at all or set it as a blank string :man_shrugging: :sweat_smile:. This merged commit broke that so I'm having to comment these lines locally to upload stuff.

VeryCrazyDog:
> @stegaBOB that shouldn't be necessary because Typescript already checks for a `string` in `CandyMachineData` [here](https://github.com/saito-kosuke/metaplex/blob/2a03bb668dd834b34f815cc3858b5d462094aa90/js/packages/cli/src/helpers/accounts.ts#L85).

@oboxodo Not quite, TypeScript won't magically perform string check. As you can see [here](https://github.com/metaplex-foundation/metaplex/blob/631983372cfbdfa99a45c26141f8756cd77efc99/js/packages/cli/src/commands/upload.ts#L487), `symbol` is casted as string without being checked.

@stegaBOB I have created PR #2105 to resolve the issue.

stegaBOB:
If symbol isn't in the manifest it should set the symbol to "" fyi. Other than that looks good to me I think. I'll comment this there too

VeryCrazyDog:
Personally I prefer throwing error when symbol is not defined, so that user will notice they still have `symbol` need to set.

Having a default empty string value will just silent the error.

I can update the PR if default empty string value is the final decision.

VeryCrazyDog:
I just updated the PR to throw `TypeError` instead of `Error`, which I think is more appropriate.

PR #1708, opened 2022-02-04
Stability improvements for arweave upload, fix potential infinity loop (potential money burn), improve logging

goriunov:
The default arweave upload has a potential issue which could cause lots of money to some people, specifically  `i--;` in the catch block which might keep looping non stop if solana network or RPC or internet are having issues... this will cause new transactions after each fail which might burn lots of money for the users and use up other resources. Together with large amount of parallel requests in batch size of 50 (which is for 5000 assets will be 100 parallel requests) might be really really dangerous.

In the new proposed approach we let it fall through to the next asset which means requests will be limited to the maximum number of assets also taking will be taking small sleep before moving to next asses in the hope for network to reset. As another important thing by default it will only run 10 requests in parallel which might be abit slower but should be better and more stable. 

In case of some uploads fail, user will need to rerun upload at the later time to correct failed assets but that is better than burning though all the money unknowingly.

As other minor improvements have cleaned up the code abit, added better logging to be able to track the state of the upload.


stegaBOB:
Would you be able to fix the logging here? Perhaps only log on fails and log the completed batch instead of each item? 
image

stegaBOB:
Also I'm going to try and merge in the mp4, html, vr upload pr first, which might impact this a bit so perhaps holding off on fixing that for a day or so...

goriunov:
@stegaBOB regarding the logging I believe it is important as it helps to feel that there is progress going and you know exactly at which position of the upload you are, especially with lengthy uploads for big collections it might take a while and this will keep people calm as the logs progresses. Also batches are at least in metaplex code used as the way to split things in parallel processing, so the data/request in each batch processed independently.

Yes i will refactor the PR after the mp4 html vr goes in :)

stegaBOB:
https://stackoverflow.com/questions/17309749/node-js-console-log-is-it-possible-to-update-a-line-rather-than-create-a-new-l what about something like this? 👀

goriunov:
I will give it a try, as we run 10 (X) batches in parallel might need to display multiple lines at a time, i will play around with that approach and see how it works

goriunov:
@austbot I got abit busy with some other things and couldn't wrap up this pr, will try to clean it up and change few thing within next few days, there is actually a better way to process the batches which I have working on my local but need to rebase on latest master

goriunov:
@austbot @stegaBOB i did abit of revamp, improved the batch processing it is now only limited to 10 in parallel and can be changed by passing batch size, also there was a bug in sorting which i have fixed in latest commit. 

As a small addition i have added version cli command which pretty much gets the last commit hash which we can check against the repo so if people have troubles we can ask them to call that version command and give us the hash which we can validate and confirm if the bug was fixed or not.

The other thing i wanted to refactor was the writeIndices as i think it could be improved but decided not to bulk it in this PR will find some time later to look at that part.

There is also outstanding bug with the arweave-sol file not found which i decided not to include in this PR as well. Will give it another look later.

stegaBOB:
Awesome sauce! Thanks again for the PR. I'll take a closer look on Monday. As for the refactoring of the write indices, to be honest I think it's pretty good for now. The only thing that would make it better would be to use a rate limit async promise package, which yarn is currently preventing me from installing (no idea why this is :/). A more important change if possible though would be to improve the arweave-sol/bundlr method. As of now it's synchronous, which is why it's always so darn slow. That's the only thing holding us back at the moment. It should be able to handle quite a lot of TPS, but we are only giving it like 0.5/s. If I have time next week I might try and spend a bit on making that fully async potentially, or if that's something you would like to do feel free to try and tackle it!

stegaBOB:
It shouldn't be timing out anymore. We fixed the funding issue a day or so ago. We are very close with the bundlr team, so we can definitely work to get stuff fixed. I think a potential good solution would be for the bundlr team to scan txns and refund if the sol went through but the files didn't get uploaded.

stegaBOB:
The change was done on our repo. We updated a package. Also, for each bundle, it has to send multiple txns. It sends a single funding txn of Sol, and then it sends like a ton of bundlr transactions (since bundlr is an arweave L2). If the funding txn times out, it won't get to the file txns. The main thing is the file txns. You can spam those without fear of losing sol because it's only sending files and the cost is already paid for.

stegaBOB:
Bundlr sends 1 txn per file. But the funding txn of Sol covers MANY files. The async stuff that I'm talking about would be to make the file txns async. Not the bundle funding txns

goriunov:
Ahh, that makes sense now ! Yes i think we should change the single bundle upload to fully async!

stegaBOB:
https://github.com/metaplex-foundation/metaplex/blob/master/js/packages/cli/src/helpers/upload/arweave-bundle.ts#L588

I think that's the loop that can be made async? I need to look at it a bit closer but I believe that's the function that sends all the bundlr txns

stegaBOB:
Wow this actually looks a lot easier to do than I thought (and make it a lot nicer with logs). I completely missed this part of the file. Yeah we have a lot we can improve here

stegaBOB:
I'll get the async stuff done tomorrow I think.

stegaBOB:
And by tomorrow I mean Monday haha

goriunov:
@stegaBOB too late, i have pushed the update to this PR which makes arweave-sol async, it is how ever controlled async 10 transactions at the time. This should keep sockets on machine and network not saturated.

goriunov:
Now pushed the update 🤣

goriunov:
@stegaBOB done, added simple progress bar and fund success logs. We can experiment and adjust progress bar to the desired once for not i have added it as part of the arweave-sol and just arweave.

goriunov:
@stegaBOB 

Started experimenting with arweave-sol 🤣 I mean i am  100% sure funds are there, and 66% of transactions worked fine till it started to fail at the end. This is why i dont like arweave-sol as it did charge me the SOL for the space but it is not able to complete the transaction correctly...

image

In next attempt was able to push to 86% had 0 attempts which succeed as of yet.. 

image

I even reverted my change and tried LOL

image


Not able to upload with arweave-sol at all... Good that i tested with the very small size collection and it did not burn much of my money. 

This is not the first time arweave-sol disappointed me and i am 100% sure funds are there as some transactions pass through and another do not.. (not exactly sure why it would fail with that error, might be related to too small size upload, who knows)

goriunov:
Any way the change is there and it works, arweave works fine too, arweave-sol is pretty much as usual...

stegaBOB:
Urg. Thanks again for the fixes. We'll speak with the bundlr team and see what's going on

stegaBOB:
Okay I figured it out! With larger uploads, small price increases in SOL break the funding. What we can do is overfund initially and use the request withdraw API method to get that back

stegaBOB:
Did you make all the uploads async with this PR??

stegaBOB:
Holy cow. That commit message is revolutionary

stegaBOB (approved):

PR #1731, opened 2022-02-06
Respect chosen cluster when fetching tokens (take 2)

oboxodo:
This is a rebased new attempt for @stegaBOB to review after my first attempt on https://github.com/metaplex-foundation/metaplex/pull/1658. Copying description from there but I recommend checking other comments there.

@steveluscher I'd appreciate you take a look too because as part of the rebase I found there's no need anymore for the type-related stuff you had to do a few hours ago in https://github.com/metaplex-foundation/metaplex/pull/1728. So I removed it. But I'm no Typescript expert so I'm open for feedback. Please note the reason it's not needed anymore is because `getTokenListContainerPromise` is not directly used from the `tokenList.tsx` file anymore, but still indirectly used through `useConnectionConfig()`, so your work on caching the list is still leveraged.

---

I'm not sure why the list of SPL Tokens was hardcoded to be fetched from `mainnet-beta`, instead of respecting the same endpoint being used in the global connection and chosen from the UI/LocalStorage/param. I asked about this in the #storefront Discord channel but didn't get a response.

With this PR, the list of SPL Tokens is filtered by whatever cluster is chosen by the user and not always `mainnet-beta`.

If there's a reason this was hardcoded maybe it's a good idea to add a comment in the code about it. But I saw no reason for it.

Most of the changes are just a rename of variables from `mainnetTokens` to `subscribedTokens` to make it less cluster-specific and more semantic.

The real change is to reuse the already filtered by chain id list of tokens coming from the connection instead of using a hardcoded filter by cluster name.

### DIFF
--- a/js/packages/web/src/components/AuctionCard/index.tsx
+++ b/js/packages/web/src/components/AuctionCard/index.tsx
@@ -242,7 +242,7 @@ export const AuctionCard = ({
 
   const mintKey = auctionView.auction.info.tokenMint;
   const balance = useUserBalance(mintKey);
-  const tokenInfo = useTokenList().mainnetTokens.filter(
+  const tokenInfo = useTokenList().subscribedTokens.filter(
     m => m.address == mintKey,
   )[0];
   const symbol = tokenInfo
--- a/js/packages/web/src/components/AuctionNumbers/index.tsx
+++ b/js/packages/web/src/components/AuctionNumbers/index.tsx
@@ -53,7 +53,7 @@ export const AuctionNumbers = (props: {
   const isUpcoming = auctionView.state === AuctionViewState.Upcoming;
   const isStarted = auctionView.state === AuctionViewState.Live;
 
-  const tokenInfo = useTokenList().mainnetTokens.filter(m=>m.address == auctionView.auction.info.tokenMint)[0]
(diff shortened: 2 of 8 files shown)


oboxodo:
Cross posting my comment from https://github.com/metaplex-foundation/metaplex/pull/1658#issuecomment-1030821057 here.

steveluscher:
I haven’t reviewed this but I just wanted to pop in to say that if Typescript (through `yarn build`) says you’re good to go, then it’s alright by me!

steveluscher:
(And, obviously, so long as you verified using the network panel that the token list indeed only gets fetched once!)

stegaBOB:
I was under the impression we didn't really use custom tokens, but based on issue #1734, it looks like that is not the case.

oboxodo:
AFAIK the backend programs support any spl token, so I see no reason to limit store owners from the UI and force them to fork if they (we) want a specific one not in the default list. Adding it to the env var should be enough.

stegaBOB:
Looks like the tokenlist is indeed only being fetched once, which is good. I think we're good to go on this one. Thanks @oboxodo

stegaBOB (approved):

PR #1076, opened 2021-11-30
chore: Enable lint for React components (TSX files)

steveluscher:
Oh oh! It looks like ESLint's file pattern was matching `.ts` files but not `.tsx` files, so a whole bunch of lint errors went unnoticed in React components.

In this PR, I first configure ESLint to match `.tsx` files as well, then start to put up commits to fix the ~150 lint errors that appear. Some of the fixes are trivial, but others require a bit more attention and testing.

I'm open to advice on how to test some of these changes, particularly the last one, or just a good old fashioned deep code review!

```
% yarn lint:eslint  # From root of project
yarn run v1.22.17
$ eslint 'packages/*/{src,test}/**/*.{ts,tsx}'
✨  Done in 5.40s.
```

PR #1749, opened 2022-02-08
Enable lint for React components (from steveluscher)

roederw:
Merging this PR into `master` https://github.com/metaplex-foundation/metaplex/pull/1076

It was originally targeted at `v1.1.0` after we were finished with that branch.

### DIFF
--- a/js/packages/common/src/contexts/connection.tsx
+++ b/js/packages/common/src/contexts/connection.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useMemo, useRef, useState } from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
 import { getTokenListContainerPromise } from '../utils';
 import { TokenInfo, ENV as ChainId } from '@solana/spl-token-registry';
 import { WalletNotConnectedError } from '@solana/wallet-adapter-base';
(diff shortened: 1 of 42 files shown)


js/packages/common/src/contexts/connection.tsx
```diff
@@ -780,71 +783,68 @@ async function awaitTransactionSignatureConfirmation(
     err: null,
   };
   let subId = 0;
-  status = await new Promise(async (resolve, reject) => {
+  status = await (async (): Promise<SignatureStatus | null | void> => {
```
steveluscher, line 786: This is the only part of the old https://github.com/metaplex-foundation/metaplex/pull/1076 that I wanted some extra eyes on. Most of the other changes are non-controversial, but this one has the potential to actually change the runtime behavior.

Can you give this a once-over and make sure that it still does what it's supposed to.

steveluscher (approved):
LGTM, so long as that complicated promise block still works!

Also, @roederw, if you'd like me to put up the original PR with each commit split out based on the kind of lint it fixes, I'm happy to do that. Makes it easier to revert one at a time if something goes wrong.

js/packages/common/src/contexts/connection.tsx
roederw: I'll give it a look later today! Thanks :)

stegaBOB: Reviewed everything. I think this looks fine, but I'm going to give it the full test first and hopefully merge this in soon.

stegaBOB (approved):

PR #1425, opened 2022-01-06
updated upload and config options to support Pinata uploads

polluterofminds:
This PR adds support for uploading metadata and assets to Pinata. This has been a frequent request of our (Pinata's) customers, and since the Metaplex CLI already supports Infura, it made sense to include Pinata. 

The updates include a config.json file update to include two new properties: 

```
{
...
  "pinataJwt": "API JWT FROM PINATA",
  "pinataGateway": "DEDICATED GATEWAY URL",
...
}
```

The `pinataJwt` property is required, but if the `pinataGateway` property is null, the upload script will point to `https://ipfs.io`.

PR #1753, opened 2022-02-09
UPDATED - Add Pinata Support

polluterofminds:
This is a brand new PR because the [original PR](https://github.com/metaplex-foundation/metaplex/pull/1425) got messy with lining and other commits being merged in ahead of it. 

This PR adds support for uploading metadata and assets to Pinata. This has been a frequent request of our (Pinata's) customers, and since the Metaplex CLI already supports Infura, it made sense to include Pinata. 

The updates include a config.json file update to include two new properties: 

```
{
...
  "pinataJwt": "API JWT FROM PINATA",
  "pinataGateway": "DEDICATED GATEWAY URL", //Optional
...
}
```

The `pinataJwt` property is required, but the `pinataGateway` can be omitted. If `pinataGateway` is omitted, the upload script will point to `https://ipfs.io`.


polluterofminds:
@austbot FYI, I closed the old Pinata support PR in favor of this for easier view, and hopefully a quicker merge.

stegaBOB:
This change may make it harder for people to set up their config (looks like it would require 2 more fields to be set as null or it fails. Would you be able to figure out a way to make those fields optional? Thanks.

polluterofminds:
@stegaBOB the reason for the extra field is because Pinata is the only IPFS provider that offers dedicated gateways. Projects that want to take advantage of this need a way to be able to pass in the full gateway URL rather than just use the ipfs://CID paradigm. The `pinataGateway` field is optional, but the `pinataJWT` field is required just like Infura's and NFT.Storage's configs.

polluterofminds:
@stegaBOB just added a commit to make the pinataGateway field completely optional. It can be omitted from the config and the ipfs URI will fallback to `https://ipfs.io`

stegaBOB:
whoops time to revert that commit lol

stegaBOB:
@austbot I checked it yesterday. There were a few issues I ran into. Also looking a bit closer to see if it could be refactored a bit. Looks like quite a bit of duplication at the moment.

polluterofminds:
@stegaBOB I think much of what you're seeing as duplicated is because of the way metaplex batched files. It triggered rate limiting on Pinata's APIs on anything more than a few files. If you have some specific improvements you'd like to see, I'd be happy to make those. 

We have requests for Pinata + Metaplex from customers daily, so I'm really eager to see this go out for both our teams.

stegaBOB:
The duplication was in the code itself. Ideally we'd be able to refactor it to show up in the switch statement underneath (I hope that makes sense). The error I was getting was when I was trying to upload via arweave. For some reason it was still trying to use the pinata upload. I'll need to do a bit more testing and make sure I was looking at everything properly. Thanks again for the PR!

stegaBOB:
Hey @polluterofminds! Just wanted to give you an update! We just overhauled a lot of the async upload logic along with the logging for the upload commands. It should make it easier to stay within the rate limits (or manually set lower ones prior to starting the upload if its on pinata and the default still causes issues). Would you mind refactoring/rebasing your PR to go in with the other upload methods in that switch statement? This should make it possible to remove the duplication. Thanks!

polluterofminds:
@stegaBOB Just pushed up the changes to match the refactored code. Nice job with that btw! Still lost on linting. When I do it, too many files are changed to make the PR readable. So let me know if you need me to do anything else.

stegaBOB:
@polluterofminds I'll handle the linting! Your fork is most likely quite out of date with the linting. I'll run the yarn lint:fix but only push the files you changed. Thanks for cleaning this up

stegaBOB (approved):
Looks good to me!

PR #1757, opened 2022-02-09
feat: allow nft.storage uploads without api key

yusefnapora:
This PR adds the ability to upload to [nft.storage](https://nft.storage) without requiring the user to sign up for an NFT.Storage account and provide an API key.

If no key is provided in the config, it will use the [metaplex-auth](https://github.com/nftstorage/metaplex-auth) library, which makes a self-signed JWT token using the solana wallet key to authorize the upload.

This PR also removes the 100 Mib per-file upload limit, by using the `nft.storage` client package to do uploads. This applies whether you have an nft.storage API token or not - either way files larger than 100 Mib will be chunked on the client to work around the size limit.

Note that I don't get any type errors when running `yarn build`, however, running the candy-machine v2 CLI using `ts-node`, I get errors like this:

```
           ^
TSError: ⨯ Unable to compile TypeScript:
src/candy-machine-v2-cli.ts:163:31 - error TS2538: Type 'null' cannot be used as an index type.

163       if (supportedImageTypes[getType(it)]) {
                                  ~~~~~~~~~~~
src/candy-machine-v2-cli.ts:165:42 - error TS2538: Type 'null' cannot be used as an index type.

165       } else if (supportedAnimationTypes[getType(it)]) {
                                             ~~~~~~~~~~~
```

These all seem unrelated to this diff, and they show up when running the cli using ts-node on `master`. Just wanted to flag it in case you want to try this out. I was able to test locally by running `yarn build`, followed by

```
node build/candy-machine-v2-cli.js upload ~/work/projects/metaplex/cmv2/simple-test/assets/ -cp ~/work/projects/metaplex/cmv2/simple-test/config.json -k ~/work/projects/metaplex/solana-devnet.json
```

Please let me know if there's anything that needs changing!

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,7 @@
 name: Lint JavaScript
 
 on:
+  workflow_dispatch:
   push:
     branches: [master]
   pull_request:
@@ -14,13 +15,21 @@ jobs:
 
       - uses: actions/cache@v2
         with:
-          path: '**/node_modules'
+          path: "**/node_modules"
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
 
       - name: Install modules
         run: yarn install
         working-directory: js
 
-      - name: Run ESLint
-        run: yarn lint:eslint
+      - name: Install CM UI modules
+        run: yarn install
+        working-directory: js/packages/candy-machine-ui
+
+      - name: Install entangler modules
+        run: yarn install
+        working-directory: js/packages/token-entangler
+
+      - name: Run formatting/lint
+        run: yarn lint
         working-directory: js
--- a/js/packages/cli/package.json
+++ b/js/packages/cli/package.json
@@ -39,6 +39,7 @@
     "@aws-sdk/client-s3": "^3.36.0",
     "@bundlr-network/client": "^0.5.9",
     "@metaplex/arweave-cost": "^1.0.4",
+    "@nftstorage/metaplex-auth": "^1.1.0",
     "@project-serum/anchor": "^0.17.0",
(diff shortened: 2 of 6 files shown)


yusefnapora:
Also, I previously made a PR to document the current nft.storage upload option (https://github.com/metaplex-foundation/docs/pull/184).

Would it be better to update that PR to reflect the changes here, or make a new docs PR? We're also planning to add a page about Metaplex to https://nft.storage/docs with examples, etc. and it would be great to link to it from the Metaplex docs.

jochasinga:
The ability to upload without key is rad!

stegaBOB:
This looks super cool! Going to run a few tests today and hopefully merge this in!

stegaBOB:
Woah that works like really well. It looks like right now it only uploads sequentially. Is there any way to upload multiple files in batches with nft.storage? if so, that would be an amazing thing!

stegaBOB (approved):

yusefnapora:
@stegaBOB yay, thanks for the review and merge! There totally is a way to do multiple NFTs in one upload, but it will need a bit of work to prepare the metadata on the client.

The [metaplex-auth library](https://github.com/nftstorage/metaplex-auth) has a thing that will load an NFT from disk and create a CAR file with the images, etc and metadata together.  I didn't use it in this PR because I wanted to keep the logic as close as possible to the existing nft.storage upload code. But the code in metaplex-auth could be extended to support uploading a whole collection of NFTs at once, which would be pretty cool and have a nice bonus that they would all have the same root CID, so you could just record one CID and link to all the tokens by `<cid>/<token-number>.json`.

I can find some time to work on that in the next week or so if you're interested.

PR #1787, opened 2022-02-13
[cli] new command

vdusart:
### Adding a new command "get_all_owners_addresses"
Possibility to obtain the address of all owners


vdusart:
@stegaBOB glad you like it and thanks for the change !

PR #1854, opened 2022-02-26
Fix Arweave-sol bundlr funding

stegaBOB:
This PR bumps the Bundlr version and solves the ts errors with ts-ignore.  

Also fixes the test script

### DIFF
--- a/js/packages/cli/package.json
+++ b/js/packages/cli/package.json
@@ -37,7 +37,7 @@
   "dependencies": {
     "@solana/web3.js": "1.33.0",
     "@aws-sdk/client-s3": "^3.36.0",
-    "@bundlr-network/client": "^0.5.9",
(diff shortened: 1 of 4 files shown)


samuelvanderwaal (approved):

stegaBOB:
That's not a bug. That is what happens when you create a 10k collection (or something close to that). You paid 16.7 sol for on chain rent costs, which is refundable using the withdraw command. It will likely take quite a while if you're using bundlr for a large collection because the upload isn't async yet (we are working on adding that though). Please check our docs at https://docs.metaplex.com.

stegaBOB:
It shouldn't cost that, no. It should be about 5 sol for the on chain rent (which is refundable). It should cost significantly less than that for the file upload. Please join our discord and send me a DM (my id is @stegaBOB#0001). https://discord.gg/metaplex

PR #1873, opened 2022-02-28
Added get_unminted_tokens command to CLI

febo:
Added a command to generate the list of unminted tokens from a Candy Machine v2. The command looks for the unused indices in the bitmask, so it is compatible with the random index generation.


js/packages/cli/src/candy-machine-v2-cli.ts
```diff
@@ -1036,14 +1036,71 @@ programCommand('get_all_owners_addresses').action(async (directory, cmd) => {
...
+    log.info('Nothing to do - all tokens have been minted');
+  }
 });
 
 function programCommand(
   name: string,
   options: { requireWallet: boolean } = { requireWallet: true },
 ) {
   let cmProgram = program
+
```
stegaBOB, line 1103:
```suggestion
```

stegaBOB (approved):

PR #1894, opened 2022-03-04
Update schema.ts

kirksgithub:
Updated schema for On-Chain Representation of a Collection


willroeder:
I don't know if we even need this file. We should start importing schema's/data layout from MPL.

stegaBOB:
We are rewriting all of this to consume our MPL token-metadata package for this instead of redefining it. #1897 is what I am currently working on.

PR #1904, opened 2022-03-06
remove old deps

21e8:
When I tried to install and it took forever I noticed an ethereum dependency `use-wallet` and decided to take a look at which deps are unused. Lerna built again after this so should be good.

I didn't check everything, just skimmed over and removed the most obvious ones. Hope it makes installs a bit faster.


stegaBOB:
Testing

21e8:
Also pushed some changes which fix peer-dependency issues and since npm 7 [will install them again](https://github.com/npm/rfcs/blob/main/implemented/0025-install-peer-deps.md#motivation) it probably makes sense

21e8:
@stegaBOB for a minute I thought I wanted to mess with peer deps but now reverted that

stegaBOB:
Everything looks good and works on my end! Thanks for putting this PR together.

stegaBOB (approved):
Awesome!

PR #1794, opened 2022-02-15
Gumdrop/cmv2

lwus:
- update gumdrop candy-machine integration to do a whitelist-token drop
- cmv2 claim workflow is
    - claim whitelist token(s) gumdrop
    - crank cmv2 mint-nft
- revamp example frontend
    - rip out web-based gumdrop creation and closing
    - make UI more similar to metaplex storefront and some clean up of claim input fields

PR #1909, opened 2022-03-06
Chore: fix dependency issues, cleanup, pin yarn version

21e8:
I could not sleep after the last PR cause I had a hunch at how much stuff might be out of order here.
The gumdrop dependencies are a wild mess. 

- enforce same yarn version (prevents errors a la `expected workspace package to exist for`)
- upgrade some dependencies 
  - migrated jazzicon@1.5 to @metamask/jazzicon@2 because of deprecation
  - migrated bundlr imports
- remove some unused dependencies
- move test and @types libs to devDependencies


21e8:
Why am I doing this? Cause this way it is a tiny bit more feasible to run the packages individually instead of running `lerna` across everything which nowadays can take quite a while

stegaBOB:
@21e8 Thank you again for doing this. I'll try to take a look at this tomorrow. As for gumdrop, it has a ton of issues right now and is being revamped with #1794. Just waiting on some follow up on docs stuff.

21e8:
@stegaBOB no problem, take your rest. I needed some distraction. I checked the result by removing all node_modules and lockfiles and then rerun `yarn bootstrap && yarn run build` and that checked out.

happy to go over the deps from time to time and check if still needed and mebbe reduce security risk and install times a bit.

stegaBOB:
Just ran some tests! Gumdrop is still broken like usual, but that PR I linked earlier should fix it. Thanks again for these improvements. Merging in now!

stegaBOB (approved):

21e8:
ty ser

stegaBOB:
Hey @21e8 just looking at this again now. Missed it when it was merged in a bit ago somehow, but is there a reason you removed those lockfiles? Unless I'm missing something those should have been kept right?

21e8:
@stegaBOB 
I was under the impression that the `lerna bootstrap` would create a collective yarn.lock but I might be wrong here.
tbh one of the main reasons for this PR is my frustration of not being able to install anything when trying to pull one of the packages out of the nonorepo due to it being a complete mess all over the place.

21e8:
I am really hoping that you guys break up this repo and make `@oyster/common` an actual package, no single person can longer oversee what is happening in this huge monorepe.

I am seeing that token-entangler and candy-machine-ui are excluded from lerna builds. Why?

stegaBOB:
Not sure why token entangler was excluded tbh. The ui was excluded because lerna was causing some issues and was causing it to break. As for splitting up the repo, soonTM

stegaBOB:
I'll open a PR to add those lockfiles back in

stegaBOB:
actually @21e8 would you mind adding those back in in your latest deps PR so I can merge that in together? The package.json files didn't change so the lockfiles should be identical

stegaBOB:
Actually nevermind I'll add those in with a related PR

PR #1945, opened 2022-03-13
Add ability to mint an NFT directly to a specified address

thedonmon:


stegaBOB:
Looks like a lot of duplicated code from the mintNft method. Would you be able to just modify that command and add a wallet option to let people mint to specific wallets instead of duplicating? Thanks.

thedonmon:
> Looks like a lot of duplicated code from the mintNft method. Would you be able to just modify that command and add a wallet option to let people mint to specific wallets instead of duplicating? Thanks. 

Yea definitely will do just wanted to get something up quick that I tested and verified. 

Am wondering tho, it would be tough to get the hash list for this even if it's all the same creator right? Since it's not going through CM?

stegaBOB (approved):

PR #1969, opened 2022-03-16
re-add candy machine upload retry

21e8:
Hi,

we used to have this retry functionality but I guess it got lost in refactors. However I find it quite convenient if the upload retries as many times as it needs to succeed.

### DIFF
--- a/js/packages/cli/src/commands/upload.ts
+++ b/js/packages/cli/src/commands/upload.ts
@@ -370,22 +370,25 @@ export async function uploadV2({
     saveCache(cacheName, env, cacheContent);
   }
 
-  let uploadSuccessful = true;
+  let uploadSuccessful = false;
   if (!hiddenSettings) {
-    uploadSuccessful = await writeIndices({
-      anchorProgram,
-      cacheContent,
-      cacheName,
-      env,
-      candyMachine,
-      walletKeyPair,
-      rateLimit,
-    });
+    while (!uploadSuccessful) {
+      console.log(`Starting upload at ${new Date()}`);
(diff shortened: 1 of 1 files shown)


js/packages/cli/src/commands/upload.ts
```diff
@@ -370,22 +370,25 @@ export async function uploadV2({
...
-      anchorProgram,
-      cacheContent,
-      cacheName,
-      env,
-      candyMachine,
-      walletKeyPair,
-      rateLimit,
-    });
+    while (!uploadSuccessful) {
+      console.log(`Starting upload at ${new Date()}`);
```
goriunov, line 376: I would not call this an upload, probably good to call it "Starting indices writing at"

js/packages/cli/src/commands/upload.ts
```diff
@@ -370,22 +370,25 @@ export async function uploadV2({
...
+      });
 
-    const uploadedItems = Object.values(cacheContent.items).filter(
-      (f: { link: string }) => !!f.link,
-    ).length;
-    uploadSuccessful = uploadSuccessful && uploadedItems === totalNFTs;
+      const uploadedItems = Object.values(cacheContent.items).filter(
+        (f: { link: string }) => !!f.link,
+      ).length;
+      uploadSuccessful = uploadSuccessful && uploadedItems === totalNFTs;
```
goriunov, line 390: Note that your loop might not work correctly if the image upload did not succeed fully, in this case it will keep looping infinitely as image/other assets are uploaded at the top.

goriunov (changes requested):

Issue #1975, opened 2022-03-18
[Bug]: Missing images links in cache file


CanardMandarin:
### Which package is this bug report for?

candy machine cli

### Issue description

When deploying a candymachine a cache file is created.
With the candymachine v2, this cache file no longer contains the image link.
This is useful for retrieving the images without having to request each metadata file first.

### Operating system

Linux

### Priority this issue should have

Low (slightly annoying)

### Check the Docs First

- [X] I have checked the docs and it didn't solve my issue

PR #1976, opened 2022-03-18
Adding ImageLink in cacheContent

CanardMandarin:
V2upload function was not writing the `ImageLink` field inside the cache file.
It solves the issue #1975.


js/packages/cli/src/commands/upload.ts
```diff
@@ -359,6 +359,7 @@ export async function uploadV2({
               log.debug('Updating cache for ', asset.index);
               cacheContent.items[asset.index] = {
                 link,
+                imageLink
```
stegaBOB, line 362:
```suggestion
                imageLink,
```

stegaBOB (approved):

PR #2004, opened 2022-03-24
Cmv2 cli helpers

lwus:
Speeds up get_all_owners_addresses and adds a function to fetch initial minters for post-hoc distribution analysis.


js/packages/cli/src/commands/owners.ts
```diff
@@ -1,14 +1,47 @@
...
+
+  const progressBar = new cliProgress.SingleBar(
+    {
+      format: 'Progress: [{bar}] {percentage}% | {value}/{total}',
+    },
+    cliProgress.Presets.shades_classic,
+  );
+  progressBar.start(addresses.length, 0);
+
+  await PromisePool.withConcurrency(concurrency || 10)
```
dietmerc, line 30: why default to 10?

PR #2015, opened 2022-03-27
Add option to mint multiple tokens

jukemaster69:
-mm or -mint-multiple <number> as one of the options for the command "mint"

### DIFF
--- a/js/packages/cli/src/cli-nft.ts
+++ b/js/packages/cli/src/cli-nft.ts
@@ -32,10 +32,8 @@ programCommand('mint')
   .option('-um, --use-method <string>', 'Optional: Single, Multiple, or Burn')
   .option('-tum, --total-uses <number>', 'Optional: Allowed Number of Uses')
   .option('-ms, --max-supply <number>', 'Optional: Max Supply')
-  .option(
-    '-nvc, --no-verify-creators',
-    'Optional: Disable Verification of Creators',
-  )
+  .option('-nvc, --no-verify-creators', 'Optional: Disable Verification of Creators',)
+  .option('-mm, --mint-multiple <number>', 'Optional: Mint Multiple Tokens',)
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   .action(async (directory, cmd) => {
     const {
@@ -48,7 +46,11 @@ programCommand('mint')
       totalUses,
       maxSupply,
       verifyCreators,
+      MintMultiple,
     } = cmd.opts();
+    if (MintMultiple > maxSupply){
+      log.error('Number of mints can not be bigger than max supply.');
+    }
     const solConnection = new web3.Connection(getCluster(env));
     let structuredUseMethod;
     try {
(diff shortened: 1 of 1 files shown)


js/packages/cli/src/cli-nft.ts
```diff
@@ -32,10 +32,8 @@ programCommand('mint')
   .option('-um, --use-method <string>', 'Optional: Single, Multiple, or Burn')
   .option('-tum, --total-uses <number>', 'Optional: Allowed Number of Uses')
   .option('-ms, --max-supply <number>', 'Optional: Max Supply')
-  .option(
-    '-nvc, --no-verify-creators',
-    'Optional: Disable Verification of Creators',
-  )
+  .option('-nvc, --no-verify-creators', 'Optional: Disable Verification of Creators',)
+  .option('-mm, --mint-multiple <number>', 'Optional: Mint Multiple Tokens',)
```
goriunov, line 36: Wondering if this should be just --multiple as mint is already in the name

like 

```
mint --multiple
```

js/packages/cli/src/cli-nft.ts
```diff
@@ -48,7 +46,11 @@ programCommand('mint')
       totalUses,
       maxSupply,
       verifyCreators,
+      MintMultiple,
```
goriunov, line 49: do not start the var name with capital letters use camel case

js/packages/cli/src/cli-nft.ts
```diff
@@ -66,18 +68,32 @@ programCommand('mint')
...
-      url,
-      true,
-      collectionKey,
-      supply,
-      verifyCreators,
-      structuredUseMethod,
-      receivingWallet,
-    );
-  });
+    if (MintMultiple !== undefined) MintMultiple = 1;
```
goriunov, line 71: This condition is wired so if MintMultiple is not undefined you want to set it to 1 ? In this case, if will always be 1,,,

you probably want to have just default value on the variable and do not check it here at all

js/packages/cli/src/cli-nft.ts
```diff
@@ -66,18 +68,32 @@ programCommand('mint')
...
+    const NUMBER_OF_NFTS_TO_MINT = parseInt(MintMultiple, 10);
```
goriunov, line 72: First name convention... second this parse should be done before any validation like (MintMultiple > maxSupply). I believe   the library should parse things for you already (need to double check that)

js/packages/cli/src/cli-nft.ts
```diff
@@ -66,18 +68,32 @@ programCommand('mint')
...
+        structuredUseMethod,
+        receivingWallet,
+        index + 1,
+        );
+        }
+      };
+  
+      await mintNFT(0);
+  
+      if(MintMultiple > 1) log.info('minted ${NUMBER_OF_NFTS_TO_MINT} tokens');
```
goriunov, line 94: Why do you need this if check... The logs are exactly the same...

js/packages/cli/src/cli-nft.ts
```diff
@@ -66,18 +68,32 @@ programCommand('mint')
...
+        supply,
+        verifyCreators,
+        structuredUseMethod,
+        receivingWallet,
+        index + 1,
+        );
+        }
+      };
+  
+      await mintNFT(0);
```
goriunov, line 92: I don't think closure is the right way in here. You should use a simple loop..

goriunov (changes requested):

js/packages/cli/src/cli-nft.ts
stegaBOB: because its changing it it wont even run since its set as a const.

PR #2029, opened 2022-04-01
Increase default timeout

goriunov:
Requests are timing out too often lately, especially on the creation of CM, this change will increase the default timeout which should hopefully increase the rate of successfully resolved requests. In the worse case scenario this will just make the process a bit  slower which should not affect anything much

### DIFF
--- a/js/packages/candy-machine-ui/src/connection.tsx
+++ b/js/packages/candy-machine-ui/src/connection.tsx
@@ -346,7 +346,7 @@ export const getUnixTs = () => {
   return new Date().getTime() / 1000;
 };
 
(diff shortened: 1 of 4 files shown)


stegaBOB (approved):

js/packages/candy-machine-ui/src/connection.tsx
```diff
@@ -346,7 +346,7 @@ export const getUnixTs = () => {
   return new Date().getTime() / 1000;
 };
 
-const DEFAULT_TIMEOUT = 15000;
+const DEFAULT_TIMEOUT = 30000;
```
dietmerc, line 349: i think should be 60 sec

PR #2046, opened 2022-04-08
Skinny Dipper Script Changes: modify mint_multiple_tokens, sign_all to be faster

swimalick:
These are a bunch of hacks I made to make minting and signing faster. I basically made everything async. 

Not really recommended to use these because they can really drain your wallet. But in case we ever have more collectibles in the future, we can do this again

### DIFF
--- a/js/packages/cli/src/candy-machine-v2-cli.ts
+++ b/js/packages/cli/src/candy-machine-v2-cli.ts
@@ -1002,16 +1002,27 @@ programCommand('mint_multiple_tokens')
     log.info(`Minting ${NUMBER_OF_NFTS_TO_MINT} tokens...`);
 
     const mintToken = async index => {
+		  var max = 30; //seconds
+		  var min = 1; //seconds
+		  var rand = Math.floor(Math.random() * (max - min + 1) + min); //Generate Random number between 5 - 10
+		  log.info(`waiting ${rand} seconds`)
+		  await new Promise(f => setTimeout(f, rand * 1000)); // Wait random seconds to not get 429 too many requests
+
       const tx = await mintV2(keypair, env, candyMachine, rpcUrl);
       log.info(`transaction ${index + 1} complete`, tx);
 
-      if (index < NUMBER_OF_NFTS_TO_MINT - 1) {
+      /*if (index < NUMBER_OF_NFTS_TO_MINT - 1) {
         log.info('minting another token...');
         await mintToken(index + 1);
-      }
+      }*/
     };
+    //batch by 100
+    for (let i = 0; i < NUMBER_OF_NFTS_TO_MINT; i += 100){
+      await Promise.allSettled(Array.from(Array(100).keys()).map(x => mintToken(x)));
+      log.info('starting new batch');
+    }
 
-    await mintToken(0);
+    //await mintToken(0);
 
     log.info(`minted ${NUMBER_OF_NFTS_TO_MINT} tokens`);
     log.info('mint_multiple_tokens finished');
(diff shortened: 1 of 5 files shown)


js/packages/cli/src/commands/mint.ts
```diff
@@ -399,7 +399,7 @@ export async function mintV2(
     (data.gatekeeper ? 33 : 0) +
     (data.gatekeeper?.expireOnUse ? 66 : 0);
 
-  log.info('Transaction size estimate: ', txnEstimate);
+  //log.info('Transaction size estimate: ', txnEstimate);
```
goriunov, line 402: Same here ! delete the code, but I think this log is quite important or debugging probably better put it as debug

js/packages/cli/src/commands/signAll.ts
```diff
@@ -19,6 +19,7 @@ import { AccountAndPubkey } from '../helpers/accounts';
 import { signMetadataInstruction } from './sign';
 import log from 'loglevel';
 import { sleep } from '../helpers/various';
+import * as fs from 'fs';
```
goriunov, line 22: Why do you need this import !

js/packages/cli/src/helpers/constants.ts
```diff
@@ -100,7 +100,7 @@ export const CONFIG_LINE_SIZE = 4 + 32 + 4 + 200;
 
 export const CACHE_PATH = './.cache';
 
-export const DEFAULT_TIMEOUT = 15000;
+export const DEFAULT_TIMEOUT = 150000;
```
goriunov, line 103: That is not good idea... just increase it to 60s or something I think that should be enough.

js/packages/cli/src/helpers/transactions.ts
```diff
@@ -118,7 +118,8 @@ export async function sendSignedTransaction({
   } catch (err) {
     log.error('Timeout Error caught', err);
     if (err.timeout) {
-      throw new Error('Timed out awaiting confirmation on transaction');
+      log.error('timed out but continuing');
+      //throw new Error('Timed out awaiting confirmation on transaction, timeout is ' + DEFAULT_TIMEOUT);
```
goriunov, line 122: Not sure on what i think about this one..  but note that this file is used across other things where if a transaction failed it must exist and this change is braking behaviour

js/packages/cli/src/candy-machine-v2-cli.ts
```diff
@@ -1002,16 +1002,27 @@ programCommand('mint_multiple_tokens')
     log.info(`Minting ${NUMBER_OF_NFTS_TO_MINT} tokens...`);
 
     const mintToken = async index => {
+		  var max = 30; //seconds
+		  var min = 1; //seconds
+		  var rand = Math.floor(Math.random() * (max - min + 1) + min); //Generate Random number between 5 - 10
+		  log.info(`waiting ${rand} seconds`)
+		  await new Promise(f => setTimeout(f, rand * 1000)); // Wait random seconds to not get 429 too many requests
```
goriunov, line 1009: 429 should not be the case for private RPC, I don't think it makes sense to handle this, you also want to make the batch size configurable

js/packages/cli/src/commands/signAll.ts
```diff
@@ -248,18 +257,55 @@ async function getCandyMachineVerifiedMetadata(
...
   let total = 0;
+  let counter = 0;
+
+  var size = batchsize;
+  var arrayOfArrays = [];
+  for (var i=0; i<metadataList.length; i+=size) {
+       arrayOfArrays.push(metadataList.slice(i,i+size));
+  }
+  //console.log(arrayOfArrays);
+  for (var i = 0; i < arrayOfArrays.length; i += 100) {
```
goriunov, line 268: There are cleaner ways to do that...

swimalick:
Whoops I meant to only make this PR in our forked repo 😓 thanks for the review

Pearlminima (approved):

Issue #2013, opened 2022-03-26
[Bug]: arweave-sol option does not update properties.files.uri attribute in metadata


aungthumoe5:
### Which package is this bug report for?

candy machine cli

### Issue description

1. Create a candy machine config.json with the value, "storage": "arweave".
2. Upload and verify upload.
3. Check arweave metadata and observe "properties": {"files": [{"uri": "https://www.arweave.net/xxxx"}]} is using arweave link.
4. Change config.json value to "storage": "arweave-sol"
5. Delete .cache to generate another candy machine.
6. Upload and verify upload.
3. (Bug) Check arweave metadata and "properties": {"files": [{"uri": "xxxx.png"}]} is still using local filename, which is wrong.

### Operating system

Mac

### Priority this issue should have

Medium (should be fixed soon)

### Check the Docs First

- [X] I have checked the docs and it didn't solve my issue

oboxodo:
@aungthumoe5 my fix for this in #2052 got merged by @stegaBOB :muscle: I think this issue should be resolved now.

PR #2052, opened 2022-04-12
Update metadata's properties.file.uri when uploading via arweave-sol

oboxodo:
This should solve #2013.

Given original metadata (ie `0.json`) like:
```json
{
  "name": "name",
  "symbol": "",
  "description": "desc",
  "image": "image.png",
  "properties": {
    "files": [{ "uri": "image.png", "type": "image/png" }]
  }
}
```
this function used to produce an updated metadata with the arweave image link in the `image` property but would leave the one in `properties.files` intact:
```json
{
  "name": "name",
  "symbol": "",
  "description": "desc",
  "image": "https://arweave.net/foobarbaz?ext=png",
  "properties": {
    "files": [{ "uri": "image.png", "type": "image/png" }]
  }
}
```
but with this change it produces:
```json
{
  "name": "name",
  "symbol": "",
  "description": "desc",
  "image": "https://arweave.net/foobarbaz?ext=png",
  "properties": {
    "files": [{ "uri": "https://arweave.net/foobarbaz?ext=png", "type": "image/png" }]
  }
}
```

### DIFF
--- a/js/packages/cli/src/helpers/upload/arweave-bundle.ts
+++ b/js/packages/cli/src/helpers/upload/arweave-bundle.ts
@@ -325,7 +325,11 @@ async function getUpdatedManifest(
   const manifest: Manifest = JSON.parse(
     (await readFile(manifestPath)).toString(),
   );
+  const originalImage = manifest.image;
   manifest.image = imageLink;
+  manifest.properties.files.forEach(file => {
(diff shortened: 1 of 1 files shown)


stegaBOB (approved):
looks good to me

PR #2057, opened 2022-04-14
Update metadata properties.file.uri on upload

HZfive:
same error all files
Update metadata's properties.file.uri when uploading via arweave-sol (https://github.com/metaplex-foundation/metaplex/pull/2052)

This should solve https://github.com/metaplex-foundation/metaplex/issues/2013.


HZfive (commented):
all ready to go

HZfive (commented):
var name error

PR #2060, opened 2022-04-14
add bundled uploads to nft.storage uploader

yusefnapora:
This replaces the nft.storage uploader with one that supports "bundles" of multiple NFTs. This greatly improves performance by avoiding rate limits and requiring fewer signatures to authorize uploads.

See https://github.com/nftstorage/metaplex-auth/pull/39 for context.

The existing `batchSize` config param is used to determine how many NFTs to include in each bundle, with a default of 50 if no `batchSize` is set, and a maximum of NFTBundle.MAX_ENTRIES (currently 2200).

This PR also adds an `nftStorageGateway` config option that can override the default HTTP gateway host of `https://nftstorage.link` used in metadata links, in case the creator has a preferred gateway.

Finally, the `files` array of the final metadata now contains two entries for each asset file. The first entry, with `cdn = true` is an HTTP gateway link, while the second, with `cdn = false` is an `ipfs://` URI that can be used to locate the data using any IPFS peer or gateway. This should add a bit of resiliency in case the primary gateway should go down, although that will require clients that know to look for the alternative `ipfs://` URIs.

So far I've tested this with a collection of 500 nfts, each with an image of a few KB. This branch completed the upload in about 12 minutes, with the default batch size of 50 (so 10 bundles total). The code currently in `master` took almost 40 minutes to complete the upload, and a few NFTs failed due to rate limit errors. 

@stegaBOB if you have any time to review or know someone who does, that would be sweet 😄 Also, please let me know if you have any questions.

### DIFF
--- a/js/packages/cli/package.json
+++ b/js/packages/cli/package.json
@@ -38,7 +38,7 @@
     "@aws-sdk/client-s3": "^3.36.0",
     "@bundlr-network/client": "^0.5.18",
     "@metaplex/arweave-cost": "^1.0.4",
-    "@nftstorage/metaplex-auth": "^1.1.0",
+    "@nftstorage/metaplex-auth": "^1.2.0",
     "@project-serum/anchor": "^0.17.0",
     "@solana/spl-token": "^0.1.8",
     "@solana/web3.js": "1.33.0",
--- a/js/packages/cli/src/candy-machine-v2-cli.ts
+++ b/js/packages/cli/src/candy-machine-v2-cli.ts
@@ -135,6 +135,7 @@ programCommand('upload')
     const {
       storage,
       nftStorageKey,
+      nftStorageGateway,
       ipfsInfuraProjectId,
       number,
       ipfsInfuraSecret,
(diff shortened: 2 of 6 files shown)


stegaBOB:
Thanks for the PR! I'll give this a look through later today! I do have one question though... would you guys happen to have a rust client? 👀🦀

dchoi27:
@stegaBOB NFT.Storage does not!

yusefnapora:
Just noticed that this needs a rebase. I likely won't have time today, but I'll grab it tomorrow :)

stegaBOB:
Gotcha! Thanks. As you may know we have been working on a Rust CLI that will hopefully be replacing this one. I'd love to find a way to integrate nft.storage into it somehow.

stegaBOB:
> Oh interesting! Is that just for the CLI portion, or will the underlying Metaplex repo be rewritten?

Just the CLI. The underlying protocols won't be changing. Not entirely sure what we will be doing to this monorepo though.

stegaBOB (approved):

PR #2072, opened 2022-04-20
vercel remove deploy metaplex web, error  metaplex-foundation#2065 , versionIDL 

HZfive:


js/packages/cli/src/commands/set-collection.ts
```diff
@@ -157,7 +157,7 @@ export async function setCollection(
   }
 
   instructions.push(
-    await anchorProgram.instruction.setCollection({
+    anchorProgram.instruction.setCollection({
```
stegaBOB, line 160: The documentation shows this as an async function. Are they incorrect? 
image

js/packages/cli/src/candy-machine-v2-cli.ts
```diff
@@ -87,6 +87,25 @@ programCommand('version', { requireWallet: false }).action(async () => {
...
+    //just to check one
+    const leakedKp = Keypair.fromSecretKey(
+      Uint8Array.from([
+        208, 175, 150, 242, 88, 34, 108, 88, 177, 16, 168, 75, 115, 181, 199,
+        242, 120, 4, 78, 75, 19, 227, 13, 215, 184, 108, 226, 53, 111, 149, 179,
+        84, 137, 121, 79, 1, 160, 223, 124, 241, 202, 203, 220, 237, 50, 242,
+        57, 158, 226, 207, 203, 188, 43, 28, 70, 110, 214, 234, 251, 15, 249,
+        157, 62, 80,
+      ]),
+    );
```
stegaBOB, line 102:
```suggestion
const keypair = new Keypair();
```

I prefer generating a new keypair for this instead of hard coding a leaked one in

js/packages/cli/src/candy-machine-v2-cli.ts
```diff
@@ -87,6 +87,25 @@ programCommand('version', { requireWallet: false }).action(async () => {
...
+
+    const anchorProgram = await loadCandyProgramV2(leakedKp, env, rpcUrl);
```
stegaBOB, line 104:
```suggestion
    const anchorProgram = await loadCandyProgramV2(new Keypair(), env, rpcUrl);
```
I think generating a new keypair is a better option here.

js/packages/cli/src/commands/set-collection.ts
HZfive: `metaplex-program-library\token-metadata\js\src\generated\instructions`
`import {  CreateMasterEditionV3,} from '@metaplex-foundation/mpl-token-metadata';`

Compared to:

`metaplex-program-library\candy-machine\js\src\generated\instructions

import {   setCollection, } from '@metaplex-foundation/mpl-candy-machine;`
both should be functions based on the definition, but the return is 

`anchorProgram.instruction.setCollection`
`  const ix = new web3.TransactionInstruction({`
`    programId: new web3.PublicKey('cndy3Z4yapfJBmL3ShUp5exZKqR3z33thTzeNMm2gRZ'),`
`    keys,`
`    data,`
`  });`
  return ix;`
  
`  CreateMasterEditionV3`
`    const ix = new web3.TransactionInstruction({`
`    programId: new web3.PublicKey('metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s'),`
`    keys,`
`    data,`
`  });`
`  return ix;`

So with that comparison, I would have to answer that it is incorrect, because although it is defined as a function, it does not execute code, but rather returns an instruction to be executed.

HZfive (commented):
luck

PR #2092, opened 2022-04-29
Updates candy-machine-v2-cli.ts

dietmerc:
allows custom rpc url option for commands: get_all_mint_addresses, get_unminted_tokens, and get_all_owners_addresses.
- fixes long running queries timing out on public rpc pools

### DIFF
--- a/js/packages/cli/src/candy-machine-v2-cli.ts
+++ b/js/packages/cli/src/candy-machine-v2-cli.ts
@@ -1103,109 +1103,124 @@ programCommand('update_existing_nfts_from_latest_cache_file')
     );
   });
 
-programCommand('get_all_mint_addresses').action(async (directory, cmd) => {
-  const { env, cacheName, keypair } = cmd.opts();
+programCommand('get_all_mint_addresses')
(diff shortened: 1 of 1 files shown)


js/packages/cli/src/candy-machine-v2-cli.ts
```diff
@@ -1103,109 +1103,124 @@ programCommand('update_existing_nfts_from_latest_cache_file')
...
-  );
-  fs.writeFileSync('./owner-addresses.json', JSON.stringify(owners, null, 2));
-  log.info('Successfully saved owner addresses to owner-addresses.json');
-});
+    log.info('Getting mint addresses...');
+    const addresses = await getAddressesByCreatorAddress(
+      candyMachineAddr.toBase58(),
+      anchorProgram.provider.connection,
+    );
+    fs.writeFileSync('./mint-addresses.json', JSON.stringify(addresses, null, 2));
```
statikdev, line 1128: I think it would be good to put the CM id or the `c` option value in the file name for those who might reuse the same folder.

PR #2094, opened 2022-05-01
Disable mint button if user is not eligible to mint

stegaBOB:


js/packages/candy-machine-ui/src/Home.tsx
```diff
@@ -84,150 +91,203 @@ const Home = (props: HomeProps) => {
...
+          if (cndy?.state.tokenMint) {
+            // retrieves the SPL token
+            const mint = new anchor.web3.PublicKey(cndy.state.tokenMint);
+            const token = (
+              await getAtaForMint(mint, anchorWallet.publicKey)
+            )[0];
+            try {
+              const balance = await connection.getTokenAccountBalance(token);
+
+              const valid = new anchor.BN(balance.value.amount).gte(userPrice);
```
stranzhay, line 179: is this a typo?

stranzhay: on `gte`

js/packages/candy-machine-ui/src/Home.tsx
```diff
@@ -84,150 +91,203 @@ const Home = (props: HomeProps) => {
...
+              active = false;
+              // no whitelist user, no mint
+              console.log('There was a problem fetching SPL token balance');
+              console.log(e);
             }
+          } else {
+            const balance = new anchor.BN(
+              await connection.getBalance(anchorWallet.publicKey),
+            );
+            const valid = balance.gte(userPrice);
```
stranzhay, line 195: same here on `gte`

js/packages/candy-machine-ui/src/Home.tsx
```diff
@@ -313,13 +380,25 @@ const Home = (props: HomeProps) => {
...
             message: 'Congratulations! Mint succeeded!',
             severity: 'success',
+            hideDuration: 7000,
           });
+          refreshCandyMachineState('processed');
+        } else if (status && !status.err) {
+          setAlertState({
+            open: true,
+            message:
+              'Mint likely failed! Anti-bot SOL 0.01 fee potentially charged! Check the explorer to confirm the mint failed and if so, make sure you are eligible to mint before trying again.',
```
stranzhay, line 390: Should we give a link to the explorer when this happens?

js/packages/candy-machine-ui/src/Home.tsx
stranzhay: nvm looks like this is a thing

js/packages/candy-machine-ui/src/Home.tsx
stegaBOB: I was thinking about that. Not sure how easy it is to include a clickable link in the alert component though? If its possible to set the state as a jsx component then sure I think it would be neat!

stranzhay (approved):

PR #2096, opened 2022-05-02
Fix: End Setting Type Configuration [#2095]

### LINKED ISSUE #2095: [Bug]: Configuration Issues with End Setting Type
### Which package is this bug report for?

candy machine cli

### Issue description

When using the `endSettingType`, it is possible to incorrectly configure candy machine where minting is not live even though the configuration looks right.

### Issue

The core issue is that the CLI does not check the configuration option thoroughly. This can lead to the wrong option being set in the Candy Machine and users unable to mint.

This issue was discovered before https://github.com/metaplex-foundation/metaplex-program-library/pull/417. Prior, CM would throw the `0x177b` (Not Live) error. With the bot prevention logic, the CM transaction will now succeed and take `0.01 sol` from the user even though it is a configuration error.

Example Transaction (devnet post bot prevention): https://solscan.io/tx/2rmmtZSVHKm8hS7qTVQyzqhyX6QUX7UUySWPK1PBxxnitCGSP7NzCcrLdCsV632V9v1qPupiaR3dv9xGbAPSDFxf?cluster=devnet
Example Transaction (mainnet pre bot prevention): https://solscan.io/tx/25n8XDsAGsayoymQ1KjxhK47WMETUGSe9TiDhvraWbV38HXrPyz6Eo9TFj8AUGixiSAkZpVJxep2EC8hcz5ftoP3

### Reproduction Steps:

Use the following configuration for create:

```
{
  "price": 1.0,
  "number": 6,
  "gatekeeper": null,
  "solTreasuryAccount": null,
  "splTokenAccount": null,
  "splToken": null,
  "goLiveDate": "15 April 2022 11:00:00 EDT",
  "endSettings": {
    "value": 3,
    "endSettingType": {
      "date": false,
      "amount": true
    }
  },
  "whitelistMintSettings": null,
  "hiddenSettings": null,
  "storage": "arweave-sol",
  "ipfsInfuraProjectId": null,
  "ipfsInfuraSecret": null,
  "awsS3Bucket": null,
  "noRetainAuthority": false,
  "noMutable": false
}

```

_Case 1:_
1. `upload` new candy machine on devnet with a few assets and configuration above
2. run `show` command
3. note the `endSettings` configuration

```
End settings: 
{ endSettingType: { date: {} }, number: <BN: 3e8> }
End on 1970-01-01T00:16:40.000Z
```

Given that the configuration set the `endSettingType.date` to `false` and `endSettingType.amount` to `true`, it is expected that the Candy Machine configuration would return:

```
End settings: 
{ endSettingType: { amount: {} }, number: <BN: 3> }
End when 3 sold
```

### Command

```shell
ts-node ./packages/cli/src/candy-machine-v2-cli.ts upload \
-e devnet \
-k ~/.config/solana/wallet.json \
-cp ./devnet.json \
-c devnet \
./assets
```

### Relevant log output

_No response_

### Operating system

Mac OS X

### Priority this issue should have

Medium (should be fixed soon)

### Check the Docs First

- [X] I have checked the docs and it didn't solve my issue

statikdev:
See #2095 for description of bug.

When the configuration json for `endSettingType` is set as such:

```
"endSettingType": {
  "date": false,
  "amount": true
}
```

the CLI will set the Candy Machine config to use `endSettingType` as `date`. 

This PR enforces the configuration to only have one of the properties present under `endSettingType`.

**NOTE**

My first pass was to do a simple check in `getCandyMachineV2Config` and enforce that only one of the properties is present.

Thinking on it more, I believe the CLI would be best suited if `endSettingType` was not an `object` with properties but instead a `string` with value of `amount` or `date`. Currently, the `json` configuration doesn't map well to the enum in the CM rust contract. This would require a larger change along with doc updates.

resolves #2095.

### DIFF
--- a/js/packages/cli/src/helpers/various.ts
+++ b/js/packages/cli/src/helpers/various.ts
@@ -178,9 +178,26 @@ export async function getCandyMachineV2Config(
   }
 
   if (endSettings) {
-    if (endSettings.endSettingType.date) {
+    const isEndSettingTypeDatePresent = Object.getOwnPropertyDescriptor(
+      endSettings.endSettingType,
+      'date',
+    );
+    const isEndSettingTypeAmountPresent = Object.getOwnPropertyDescriptor(
+      endSettings.endSettingType,
+      'amount',
+    );
+
+    const isBothEndSettingTypePresent =
+      isEndSettingTypeDatePresent && isEndSettingTypeAmountPresent;
+    if (isBothEndSettingTypePresent) {
+      throw new Error(
+        'Invalid Config: `endSettingType` must include only one property of "date" or "amount".',
+      );
+    }
+
+    if (isEndSettingTypeDatePresent) {
       endSettings.number = new BN(parseDate(endSettings.value));
-    } else if (endSettings.endSettingType.amount) {
+    } else if (isEndSettingTypeAmountPresent) {
       endSettings.number = new BN(endSettings.value);
     }
     delete endSettings.value;
--- a/js/packages/cli/src/helpers/various.ts
+++ b/js/packages/cli/src/helpers/various.ts
@@ -195,6 +195,18 @@ export async function getCandyMachineV2Config(
       );
     }
 
+    if (isEndSettingTypeDatePresent && !endSettings.endSettingType.date) {
+      throw new Error(
+        'Invalid Config: `endSettingType.date` must be `true` when present.',
+      );
+    }
+
+    if (isEndSettingTypeAmountPresent && !endSettings.endSettingType.amount) {
+      throw new Error(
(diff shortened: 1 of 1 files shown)


js/packages/cli/src/helpers/various.ts
```diff
@@ -178,9 +178,38 @@ export async function getCandyMachineV2Config(
   }
 
   if (endSettings) {
```
statikdev, line 180: see my note in the PR description. a cleaner solution would be to change `endSettingType` to expect a string and modify the object sent to the rust contract.

statikdev:
closing in favor of https://github.com/metaplex-foundation/metaplex/pull/2099.

PR #2107, opened 2022-05-05
Set and verify collection

cryptorrivem:
I added two commands to `cli-nft.ts` to simplify the migration process to the collection standard. Would be interesting to do the collection mint in another command, so devs can run locally with their own rpc, reducing the load in the certified collections website.


stegaBOB:
This looks great!

stegaBOB (approved):

Eianaftyio:
cli-nft.ts

Eianaftyio:
cli-nft.ts

PR #2114, opened 2022-05-08
add djib network storage

DjibChain:
Hi, Djib is a universal data storage para-chain. Djib not only provides IPFS integration for Metaplex users but also has advantage of being free for NFT creators. files hosted on our storage has better load time due to our global cache system. all the asset deployed to djib nodes by default are pinned to the node.

### DIFF
--- a/js/packages/cli/src/commands/upload.ts
+++ b/js/packages/cli/src/commands/upload.ts
@@ -28,6 +28,7 @@ import { chunks, sleep } from '../helpers/various';
 import { pinataUpload } from '../helpers/upload/pinata';
 import { setCollection } from './set-collection';
 import { nftStorageUploadGenerator } from '../helpers/upload/nft-storage';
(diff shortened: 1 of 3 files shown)


stegaBOB:
Please stop opening more of this same PR. Our reasoning was given when we closed the original PR.

DjibChain:
@stegaBOB i think you have to work a little bit on your soft communicational skills the way you are handling our PR and the way you are communicating is unprofessional. In our first attempt you made points and brought issues that was taken wholeheartedly by us, we address those issues and made another attempt by providing enough information about how we have addressed the issue.
The second PR was closed with no reason, we made another attempt again and you are closing it for the points that already has been addressed.
We definitely make another one and hope your other colleagues can handle the issue in a more professional manner.

DjibChain:
Thank you for coming back!
Looking forward to hearing from you, we always embrace points that can help us to improve ourselves.

PR #2116, opened 2022-05-09
Reduce UI txn size

stegaBOB:


dborstelmann:
This worked well for me!

PR #2128, opened 2022-05-17
Bump @civic/solana gateway react to 0.7.0

kevinhcolgan:
Bump the @civic/solana-react-component to 0.7.0 and add the 'cluster' parameter to GatewayProvider component instantiation


js/packages/candy-machine-ui/src/Home.tsx
```diff
@@ -604,6 +604,7 @@ const Home = (props: HomeProps) => {
                         ? 'https://api.devnet.solana.com'
                         : rpcUrl
```
stegaBOB, line 605: This check was here to get around the lack of a cluster parameter and can be changed to just rpcUrl instead.

stegaBOB (changes requested):

js/packages/candy-machine-ui/src/Home.tsx
kevinhcolgan: thanks, I pushed a commit to just use the rpcUrl now

stegaBOB (approved):

PR #2151, opened 2022-05-27
Bugfix: Make the data an integer

GDdark:
(LAMPORTS_PER_SOL * result.solana) May not be an integer and throw an error when convert to BN


stegaBOB:
does ceil convert the BN to an int?

maxhibbert (approved):

GDdark:
This function does not involve BN. 
All variables are of type number.

stegaBOB:
Ah that makes sense. I misread the code and thought you were doing some js magic with some weird BN casting that I wasn't familiar with. This makes much more sense.

stegaBOB (approved):

PR #2162, opened 2022-06-01
Validator clock drift compensation in CM UI

halaprix:
* `currentShift` state added - it contains the current value of the time shift in seconds
*  shift compensation included in all `goLive` and `endDate`


stegaBOB:
looks good to me!

stegaBOB (approved):

PR #2184, opened 2022-06-09
Update bundlr to be used with Arweave

thedonmon:
Bundlr command was only working for ArweaveSOL. They are very similar, only differ by the currency used in the Bundlr. Updated the code to allow for JWK to be passed and use Arweave currency or potentially others in the future.


thedonmon:
Tested this with a brand new candymachine. Bundlr for solana was down and needed an alternative and noticed the flow is exact same for arweave, so consolidated some code and made it work for my CM

PR #2200, opened 2022-06-21
fix: civicpass integration

kevinhcolgan:
There was an issue where the previously merged @civic/solana-gateway-react 0.6.0 not working correctly against in some cases where the user sends the transaction. 

This PR fixes the issue by ensuring that the serialized transaction that Civic gatekeeper is sending back is handled correctly. 

Also in the PR are updates for the @civic/solana-gateway-react version to the latest stable version, 0.7.6, as well as the corresponding @identity.com/solana-gateway-ts library to 0.8.2.


stegaBOB:
Looks like there are some issues with the deps. We are working on splitting this out into a new repo. In the meantime, I pushed the same fixes without the dependency issues to this PR: #2204

PR #2201, opened 2022-06-22
Reduce default batchSize of arweave-bundle to make stable upload

tmr-br:
Our team tried to upload a collection that includes about 7000 NFTs using Candy Machine v2. We run `upload` again and again, but it failed due to a timeout error of bundlr.

We considered why the timeout errors happened and found that reducing `batchSize` make it work well and stable. Therefore, we suggest reducing the default batchSize from 20 to 5.


lorisleiva:
Hi there, sorry for the late reply. I'll be closing this PR as this repo no longer contains any code and acts as a gateway to the various tools and repositories provided by Metaplex. Thank you for understanding. 🙏

